PR #8, opened 2017-10-22
Try cacheing Ruby gems and Node modules in Travis

davidrunger:


davidrunger:
Well, I guess I'll merge and we'll have to find out later if this worked or not. :-p :)

davidrunger:
Bah, I don't think it worked. I'll give it a little while longer, though, before I make a call.

PR #29, opened 2017-10-29
Setup `sudo: false` in Travis config

davidrunger:
Supposedly this helps with bundler cacheing?
https://bibwild.wordpress.com/2015/04/08/simple-config-for-faster-travis-ruby-builds/


davidrunger:
Seems to have worked! Or maybe it was #28 that worked.

PR #39, opened 2017-11-18
Update flex CSS rules/usage per basscss

davidrunger:


app/javascript/home/home.vue
```diff
@@ -22,7 +22,7 @@ div
           span.ptb-1 Resume
         a.nav-link(href='#contact')
           span.ptb-1 Contact
-    a.down-arrow-container.flex-grow-0(href='#about')
```
davidrunger: this didn't seem to be actually doing anything

PR #48, opened 2017-11-18
Import routes in `customized_vue.js` (not individual packs)

davidrunger:
This fixes an error in production.

The problem is that $routes was undefined on bootstrap instances because `Routes` hadn't been loaded yet at the time that `customized_vue.js` was loaded.

This PR fixes that bug, and DRYs up the code a bit, too.


davidrunger:
I don't understand this PR description, but ok. :-p

PR #103, opened 2017-11-26
Upload JavaScript source maps to Rollbar after assets:precompile

davidrunger:
This will allow us to have source maps in Rollbar for JavaScript errors.


davidrunger:
I think that I used `RestClient` rather than `HTTParty` here because `HTTParty` did not support multi-part file posting at the time (which we need to post the sourcemaps to Rollbar). But now it does.

TODO:
> Leverage the fact that HTTParty now has multipart support to drop RestClient dependency

PR #104, opened 2017-11-26
Remove accidentally-left-in binding.pry

davidrunger:
Woops! :upside_down_smiley_face:


davidrunger:
I have a check for this sort of mistake built into my shell command that initiates a pull request, but it only looks at .rb, .js, .coffee, .haml, .html, and .erb files, so it missed this because it was in a .rake file. :sob: I guess I should add .rake to the list...

PR #157, opened 2017-12-13
Exit block with `next` not `break`

davidrunger:
Apparently `break` is [specific to `#each`][1] or something?

[1]: https://stackoverflow.com/a/9795964/4009384


davidrunger:
Fixes https://rollbar.com/davidjrunger/davidrunger/items/47/

PR #161, opened 2017-12-16
Don't save Requests for New Relic pings

davidrunger:
I've setup New Relic Synthetics pinging. The New Relic Synthetics agent appears to be Chrome 36 (no longer NewRelicPinger), so we are currently logging these requests.

It's kind of interesting to see the requests from all over the USA, but it's misleading/confusing, since I mostly mean for requests to gather ideally just requests from humans, or at least bots that I don't already know about and expect regularly (such as the New Relic Synthetics requests).

I've added an ENV var check, in case I do want to look at the New Relic requests every once in a while. I'll be able to easily toggle the ENV var on and off.

### DIFF
--- a/app/models/request.rb
+++ b/app/models/request.rb
@@ -26,14 +26,17 @@
 #
 
 class Request < ApplicationRecord
+  # error class for Rollbar logging
+  class CreateRequestError < StandardError ; end
+
   belongs_to :user, optional: true
 
   validates :url, :handler, :method, :format, :ip, :requested_at, presence: true
   validates :bot, inclusion: [true, false]
 
   scope :human, -> { where(bot: false) }
-  scope :recent, ->(time_period = 1.day) do
+  scope :recent, (lambda do |time_period = 1.day|
     where('requests.requested_at > ?', time_period.ago)
-  end
+  end)
   scope :ordered, ->() { order('requests.requested_at') }
 end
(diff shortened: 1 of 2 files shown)


app/models/request.rb
```diff
@@ -26,14 +26,17 @@
...
+  class CreateRequestError < StandardError ; end
+
   belongs_to :user, optional: true
 
   validates :url, :handler, :method, :format, :ip, :requested_at, presence: true
   validates :bot, inclusion: [true, false]
 
   scope :human, -> { where(bot: false) }
-  scope :recent, ->(time_period = 1.day) do
+  scope :recent, (lambda do |time_period = 1.day|
```
davidrunger: for Rubocop

config/initializers/request_logging.rb
```diff
@@ -19,45 +10,41 @@ class CreateRequestError < StandardError ; end
 
   next if stashed_data['admin'] == true
 
-  user_agent = stashed_data['user_agent']
```
davidrunger: there's no need to extract this local variable anymore, since `stashed_data['user_agent']` is only being used once now. previously it was being used twice (once to check if it's new relic, and then again to log it)

config/initializers/request_logging.rb
```diff
@@ -19,45 +10,41 @@ class CreateRequestError < StandardError ; end
...
+  params = stashed_data['params']
```
davidrunger: inversely to https://github.com/davidrunger/david_runger/pull/161/files#r157351140, we are now using `stashed_data['params']` twice, so keep it DRY via a local variable

PR #162, opened 2017-12-17
Attempt to run StatsD on Heroku web dyno

davidrunger:
This is an experiment. We'll see if it works. If it works, then I can work next on actually piping the StatsD output to a service that I can use to view/process that data.

**Commits:**

1. Add StatsD daemon JavaScript library. Installed from GitHub since this fix hasn't been released yet https://github.com/etsy/statsd/pull/593.
2. Add temp testing StatsD deamon configuration
3. Add Shopify's StatsD Ruby library
4. Configure StatsD client
5. Add Procfile & Procfile.multi to run multiple Heroku processes. Huge shoutout to [James Aylett](https://tartarus.org/james/diary/2013/04/18/running-statsd-on-heroku) for this hack/idea! Why aren't more people doing this? Well, maybe we should see if it works first. :-p
6. Use StatsD to track success/failure of Grocery store creation

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,3 +1,4 @@
+statsd-config.js
 /node_modules
 /public/packs
 /app/javascript/rails_assets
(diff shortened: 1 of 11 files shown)


davidrunger:
Come the think of it, this `Procfile.multi` hack might also work to run Sidekiq?

PR #209, opened 2017-12-25
Focus item name input when triggering editing

davidrunger:
Previously, I was using `autofocus` for this, but the problem is that Chrome only seems to respect `autofocus` once; subsequent double- clicking of other inputs (or even the same input, IIRC) would result in an unfocused input.

This PR moves us to managing focus manually so that the input will _always_ be focused when editing is triggered.

### DIFF
--- a/app/javascript/groceries/components/item.vue
+++ b/app/javascript/groceries/components/item.vue
@@ -4,14 +4,14 @@
       span.increment.h2.js-link.olive(@click='setNeeded(item, item.needed + 1)' title='Increment') +
       span.decrement.h2.pl1.pr1.js-link.red(@click='setNeeded(item, item.needed - 1)' title='Decrement') &ndash;
       input(
-        v-if='editingName'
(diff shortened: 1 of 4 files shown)


davidrunger:
Although I have much ❤️  for `vue-test-utils`, I did run into some issues with it in this PR:
1. I ran into this bug, which is fixed on the `dev` branch and will be released in `1.0.0-beta.9`, but since that is not available yet via npm/yarn, I've upgraded to track the `dev` branch via GitHub
2. When I was using the `hasStyle` matcher, to check for the `input` to `hasStyle('display', 'none')`, the elements were getting added to the DOM and then not cleaned up (in other words, the elements were appearing at the bottom of my mocha test output in the browser). I imagine that the only way for the `hasStyle` matcher to actually work is to render the element into the DOM, but it's a little unfortunate that the element isn't cleaned up afterward. I wonder if there is a reason for this?

PR #264, opened 2017-12-29
Refactor Api::TextMessagesController#create

davidrunger:
The need for this refactor was largely highlighted by the excellent [RubyCritic][1] gem. Thanks, RubyCritic! 😄 👍 

[1]: https://github.com/whitesmith/rubycritic

Also:
1. Add Pundit authorize check to Api::TextMessagesController#create
1. Add SmsRecordPolicy (with #create? check)
1. Generate Pundit ApplicationPolicy via `rails g pundit:install` (and then tweaking for Rubocop compliance).
1. Add pundit gem to Gemfile & Gemfile.lock
1. Alphabetize pghero correctly in Gemfile
1. Whitelist Gemfile methods to be called without parens
1. Add fail to Rubocop list of methods that may be called w/o parens


app/controllers/api/text_messages_controller.rb
```diff
@@ -9,6 +9,7 @@ def create
     message_type = text_message_params['message_type']
     message_params = text_message_params['message_params']
     message = __send__("#{message_type}_message", message_params)
+    authorize(SmsRecord)
```
davidrunger: usually I would add this at the top of the method; it just got added here basicall because I was breaking it off as a smaller commit from a larger initial commit, so `git` worked it out that way. the above lines will be deleted in a subsequent commit.

davidrunger:
I just ran `rubycritic` on the changed code, and happily this refactor did indeed change the file from a `B` to an `A` in `rubycritic`'s eyes ... so hopefully that means that I'm doing something right! :) 📦

PR #270, opened 2017-12-30
Boot multiple Puma worker processes

davidrunger:
The default number of workers will be 3. I've also just now set the corresponding `ENV` variable (`'WEB_CONCURRENCY'`) to `3` on production (as well as setting the `'RAILS_MAX_THREADS'` env variable to its code- specified default value of `5`).

### DIFF
--- a/config/puma.rb
+++ b/config/puma.rb
@@ -4,16 +4,16 @@
 # the maximum value specified for Puma. Default is set to 5 threads for minimum
 # and maximum; this matches the default thread size of Active Record.
 #
-threads_count = ENV.fetch("RAILS_MAX_THREADS") { 5 }
-threads threads_count, threads_count
+threads_count = ENV.fetch('RAILS_MAX_THREADS') { 5 }
+threads(threads_count, threads_count)
 
 # Specifies the `port` that Puma will listen on to receive requests; default is 3000.
 #
-port        ENV.fetch("PORT") { 3000 }
+port(ENV.fetch('PORT') { 3000 })
 
 # Specifies the `environment` that Puma will run in.
 #
-environment ENV.fetch("RAILS_ENV") { "development" }
+environment(ENV.fetch('RAILS_ENV') { 'development' })
 
 # Specifies the number of `workers` to boot in clustered mode.
 # Workers are forked webserver processes. If using threads and workers together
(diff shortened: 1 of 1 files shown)


config/puma.rb
```diff
@@ -4,24 +4,24 @@
...
+environment(ENV.fetch('RAILS_ENV') { 'development' })
 
 # Specifies the number of `workers` to boot in clustered mode.
 # Workers are forked webserver processes. If using threads and workers together
 # the concurrency of the application would be max `threads` * `workers`.
 # Workers do not work on JRuby or Windows (both of which do not support
 # processes).
 #
-# workers ENV.fetch("WEB_CONCURRENCY") { 2 }
+workers(ENV.fetch('WEB_CONCURRENCY') { 3 })
```
davidrunger: uncommenting this line is really the main change in this PR (and also using 2 workers instead of the commented-out default of 2)

davidrunger: I should explain how I arrived at a value of `3` workers.

I am on Heroku hobby dynos, so 512MB of memory. Looking at my Heroku "Metrics" tab, it appears that the maximum memory usage that I was attaining previously was ~140MB. (Some of that memory is actually being used by StatsD rather than the Rails server process, so the actual Rails server process is using less than 140MB, maybe closer to 130MB if StatsD is using 10MB, but for a more conservative estimate I just ignored that and assumed that all of the memory usage is from Rails.)

So, it's pretty simple: `3 * 140MB == 420MB`, which leaves a fairly safe padding of 92MB of memory before we'd start running out of memory. Additionally, due to some memory being in shared in common between the forked Puma worker processes, we can expect that the used memory will actually be less than `3 * 140MB` (with the exact difference depending on just how much memory is able to be shared between the processes over time without being copy-on-write'd).

davidrunger:
Here we can begin to see the impact of this change on memory usage. For now, memory usage appears to be reaching only ~300MB, well within the 512MB hobby size dyno limit.

image

davidrunger:
Considering that I probably don't get too many concurrent requests to https://www.davidrunger.com, this change probably won't have a ton of overall positive impact, but hey, I wanted to do it, anyway. :)

I might consider dropping the worker count down to `2` if I ever add a Sidekiq process to my web dyno (using Foreman, as I did when I added a StatsD process to my web dyno).

PR #286, opened 2018-01-01
Add test comment

davidrunger:


davidrunger:
I am just trying to trigger a Travis CI build with this branch and seeing if it works. Travis is now admitting that they have a problem, which is really what I wanted. Closing now.

PR #318, opened 2018-01-12
Use no-sandbox option when running headless Chrome on Travis

davidrunger:
Sad, but it begins to seem that this might be necessary, possibly caused by the Ubuntu image upgrade [described here][1], particularly the Chrome update.

[1]: https://docs.travis-ci.com/user/build-environment-updates/2017-12-12/


davidrunger:
Well, hey, at least it did work! That's something! 😄  👍

PR #461, opened 2018-03-05
Add .git to webpack URL for package.json/yarn.lock

davidrunger:
We're having `yarn install` issues on Travis CI. Maybe this will help?


davidrunger:
Seems to have done the trick! Maybe I should add a note to this Stackoverflow answer that I had used, which does not include this evidently sometimes necessary `.git` extension ... https://stackoverflow.com/a/43636577/4009384

PR #464, opened 2018-03-06
Remove stylelint-config-standard (temporarily)

davidrunger:
... while waiting for stylelint webpack v4 compatibility.


davidrunger:
Nevermind ... I just removed `stylelint-webpack-plugin` in 65366db ; we still have `stylelint` in general, though.

PR #525, opened 2018-03-23
Yarn upgrade

davidrunger:
I basically just yarn `yarn upgrade`. `yarn check` was then producing 3 errors. Next: `rm -rf node_modules` `yarn`. I think this got us down to one error, which is mocha-webpack had a peer dependency of webpack 2 or 3 (not 4). => `yarn add -D mocha-webpack@2.0.0-beta.0` => no more errors in `yarn check` (though we still have 3 warnings, which I will probably make at least a small effort to fix, next).

The reason that I did this in the first place is because I accidentally merged #522 while dependabot was rebasing it. So I wanted make sure that my `yarn.lock` hadn't gotten into a weird state. I did not expect such a large diff; I hardly expected any diff at all. I guess dependabot does not update secondary dependencies, only primary dependencies?

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
     "expect": "^22.4.3",
     "mocha": "^5.0.5",
     "mocha-headless-chrome": "^1.8.2",
-    "mocha-webpack": "^1.1.0",
+    "mocha-webpack": "2.0.0-beta.0",
     "postcss": "^6.0.21",
     "postcss-html": "^0.15.0",
     "postcss-safe-parser": "^3.0.1",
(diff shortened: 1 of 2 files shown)


davidrunger:
Oh, also, one note: when i did `yarn` (i.e. `yarn install`), i got an error -- something (maybe `node-pry-gyp` or whatever?) was trying to use a `python2` command (weird?), whereas i had python set to 3.4.3 via pyenv. so i temporarily did `pyenv local 2.7.10` and it made that error go away. then i `rm .python-version` to go back to version 3.4.3.

PR #547, opened 2018-04-01
Move logged-in header to sidebar

davidrunger:
It's kind of a long story why I even started down this road in the first place. Suffice it to say, though, that I think that this is better. The left-hand panel is more the "controls" of the app (choosing a store, and now also configuring user settings & logging out); it didn't make as much sense to have the logged-in header mixed in among the main panel of the grocery store and items.

## After

**Unexpanded:**

**Expanded:**

### DIFF
--- a/app/javascript/shared/customized_vue.js
+++ b/app/javascript/shared/customized_vue.js
@@ -1,19 +1,22 @@
 import axios from 'axios';
 // don't (explicitly) load  element-ui button.css, because it's included in dropdown.css
+import 'element-ui/lib/theme-chalk/base.css';
+import 'element-ui/lib/theme-chalk/button.css';
 import 'element-ui/lib/theme-chalk/card.css';
-import 'element-ui/lib/theme-chalk/dropdown.css';
-import 'element-ui/lib/theme-chalk/dropdown-menu.css';
 import 'element-ui/lib/theme-chalk/dropdown-item.css';
 import 'element-ui/lib/theme-chalk/icon.css';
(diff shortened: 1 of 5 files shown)


spec/javascript/groceries/components/sidebar.spec.js
```diff
@@ -22,9 +22,12 @@ describe('Sidebar', function () { // eslint-disable-line func-names
   beforeEach(() => {
     item = { id: 48, name: 'bananas', needed: 0 };
     groceryStore = { id: 23, name: 'Costco', items: [item] };
-    bootstrap = { stores: [groceryStore] };
+    bootstrap = {
+      stores: [groceryStore],
+      current_user: { id: 1, email: 'davidjrunger@gmail.com' },
```
davidrunger: this test now needs `current_user` in the bootstrap data since the Sidebar now renders current user information via the LoggedInHeader

app/javascript/shared/customized_vue.js
```diff
@@ -1,19 +1,22 @@
 import axios from 'axios';
 // don't (explicitly) load  element-ui button.css, because it's included in dropdown.css
+import 'element-ui/lib/theme-chalk/base.css';
```
davidrunger: it took me a while to figure out why the ElementUI menu transitions weren't working (the menu was just immediately appearing/disappearing, rather than expanding/collapsing up and down). the problem was that I wasn't importing this `base.css` from element ui's chalk theme (which, i assume, probably defines Element UI's generic CSS transitions).

PR #578, opened 2018-04-15
Get stylelint working again

davidrunger:
I had removed `stylelint-webpack-plugin` in 65366db / #436 because it was not yet compatibile with Webpack 4. But now it is! :tada:

### DIFF
--- a/app/javascript/home/home.vue
+++ b/app/javascript/home/home.vue
@@ -448,18 +448,18 @@ i[class^=devicon-] {
         color: white;
       }
 
-      &:not(.active) span {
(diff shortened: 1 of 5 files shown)


app/javascript/home/home.vue
```diff
@@ -448,18 +448,18 @@ i[class^=devicon-] {
         color: white;
       }
 
-      &:not(.active) span {
-        border-bottom-color: rgba($white-dark, 0);
+      span {
```
davidrunger: (this change was requested by stylelint, related to the ordering of specificity of selectors)

PR #606, opened 2018-04-29
Switch home font from Karla to Nunito

davidrunger:
I think it looks better. The Karla was a bit noisy. The Nunito is very simple; the reader focuses on the content, not the font.


davidrunger:
It's worth noting that Nunito is about twice as heavy as Karla (which is doubled again, since we are downloading both 400-weight and a 700-weight versions).

**Karla:**
image

**Nunito:**
image

I think it's worth it, though.

PR #608, opened 2018-04-29
Remove Serpent.js project

davidrunger:
This Serpent.js game still is a bit janky, and five projects was too many. Even four projects might still be too many. I need to re-write the projects section so that it contains more of an expansion-based UI (and so that I can call out more minor projects vs. more major projects); then I can think about adding back some of my minor projects.


davidrunger:
* branch name should be `remove-serpent`, not `remote-serpent`

PR #640, opened 2018-05-05
Source jscodeshift and nomnom from GitHub to remove deprec. warnings

davidrunger:
JavaScript is at it again! Also, these jscodeshift maintainers seem to be not really too concerned/responsive...


davidrunger:
Merged #641 instead.

PR #653, opened 2018-05-07
Bump node to 10.0.0 and yarn to 1.6.0

davidrunger:
I was going to do these bumps in separate PRs, but `yarn install` is failing on Travis when just bumping `node` alone, so let's see if also bumping yarn at the same time helps.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,13 +16,13 @@ before_install:
   - gem update --system
   - gem install bundler -v 1.16.1
   - . $HOME/.nvm/nvm.sh
-  - nvm install 8.9.3
-  - nvm use 8.9.3
+  - nvm install 10.0.0
+  - nvm use 10.0.0
   - curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.5.1
   - export PATH="$HOME/.yarn/bin:$PATH"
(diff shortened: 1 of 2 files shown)


davidrunger:
Actually looking at the error in Travis that I got when just bumping `node` by itself, it was probably just a random network error or some issue on `yarn`'s end (rather than an issue of trying to bump `node` without bumping `yarn`), but whatever.

The error:
```
$ yarn install
yarn install v1.5.1
(node:7044) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
[1/5] Validating package.json...
[2/5] Resolving packages...
[3/5] Fetching packages...
error An unexpected error occurred: "https://registry.yarnpkg.com/async-validator/-/async-validator-1.8.2.tgz: Request failed \"503 Service Unavailable\"".
info If you think this is a bug, please open a bug report with the information provided in "/home/travis/build/davidrunger/david_runger/yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
The command "yarn install" exited with 1.
```

PR #665, opened 2018-05-13
Add MVP "Log" app (for tracking weight, exercise, etc)

davidrunger:
(Currently we are only tracking weight)


app/javascript/log/log.vue
```diff
@@ -0,0 +1,84 @@
...
+  methods: {
+    postWeightLog() {
+      if (this.formstate.$invalid) return;
+
+      const payload = {
+        weight_log: {
+          weight: this.newWeightLogWeight,
+        },
+      };
+      this.$http.post(this.$routes.api_weight_logs_path(), payload).then(() => {
```
davidrunger: this should / probably will be moved into a Vuex action soon enough

PR #716, opened 2018-05-26
Disable log submit buttons while data is being posted

davidrunger:
This should prevent accidentally double-submitting any values by hitting enter multiple times in relatively rapid succession or by clicking a submit button multiple times in relatively rapid succession.

### DIFF
--- a/app/javascript/log/components/weight_section.vue
+++ b/app/javascript/log/components/weight_section.vue
@@ -12,7 +12,11 @@ div
             name='newWeightLogWeight'
             required
           )
-        el-button.flex-0(:disabled='formstate.$invalid') Add
+        el-input.flex-0(
+          type='submit'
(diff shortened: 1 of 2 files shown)


app/javascript/log/components/weight_section.vue
```diff
@@ -12,7 +12,11 @@ div
             name='newWeightLogWeight'
             required
           )
-        el-button.flex-0(:disabled='formstate.$invalid') Add
+        el-input.flex-0(
```
davidrunger: changing this from `el-button` to `el-input(type='submit')` is an unrelated change; i noticed that clicking the `el-button` button did not actually submit the form, though (which is interesting, since i thought that the default behavior of a `button` within a `form` is to submit the form? i didn't look into it, though.).

PR #1161, opened 2018-12-08
Bump vue, vue-template-compiler from 2.1.17 to 2.1.18

davidrunger:
This combines #1157 and #1158 together. I figure these are supposed to be updated at the same time, and maybe because those PRs attempt to update them separately is why the builds are failing on those PRs.

PR #1162, opened 2018-12-09
Add test coverage for groceries.vue

davidrunger:
This addresses the missing spec coverage called out here: https://github.com/davidrunger/david_runger/pull/1161#issuecomment-445485276

### DIFF
--- a/app/javascript/groceries/components/sidebar.vue
+++ b/app/javascript/groceries/components/sidebar.vue
@@ -36,6 +36,8 @@ import _ from 'lodash';
 import LoggedInHeader from './logged_in_header.vue';
 
 export default {
+  name: 'Sidebar',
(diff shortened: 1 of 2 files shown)


app/javascript/groceries/components/sidebar.vue
```diff
@@ -36,6 +36,8 @@ import _ from 'lodash';
 import LoggedInHeader from './logged_in_header.vue';
 
 export default {
+  name: 'Sidebar',
```
davidrunger: adding this `name` property was necessary in order for this test to pass:
```js
it('renders the sidebar', () => {
   expect(wrapper.findAll({ name: 'Sidebar' }).length).toEqual(1);
 });
```

PR #1183, opened 2018-12-16
Improve performance of logs#index action

davidrunger:
See commit messages for details.
1. Add performance profiling gems (13d4708)
2. Eager-load Log associations (7371d90)
3. Serialize raw database timestamp rather than parsing into a `Time` object (7739bae)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -48,10 +48,17 @@ group :development do
   gem 'annotate'
   gem 'better_errors'
   gem 'binding_of_caller'
+  # Provides flamegraphs for rack-mini-profiler.
+  gem 'flamegraph'
   gem 'listen', '>= 3.0.5', '< 3.2'
+  # Performance profiling. Should be listed after `pg` (and `rails`?) gems to get database performance
+  # analysis.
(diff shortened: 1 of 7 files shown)


davidrunger:
# Production after

```
```

**Average:  seconds**

Note that the performance improvement on production is somewhat washed out by the network round trip from where I am (San Diego, CA) to where my Heroku web servers are (Virginia, USA, I think).

davidrunger:
# Production after ... after

```
❯ for i in {1..5}; curl -s -w "%{time_total}\n" -o /dev/null 'https://www.davidrunger.com/logs' -H 'Connection: keep-alive' [...]
0.828126
0.811444
0.759502
0.742938
0.829825
```

**Average: 0.794 seconds**

I restarted dynos (by scaling down to 0 and back up to 1 via `heroku ps:scale`) and response times seemed to improve.

I'm thinking that with network variance and other issues maybe measuring on production just isn't gonna be reliable/meaningful.

davidrunger:
Here we can see better NewRelic performance now, too:

image

PR #1189, opened 2018-12-18
Fix webpacker compilation after 4.0.0-rc.2 upgrade

davidrunger:
Apparently we are now minifying via some other means than `uglifyjs-webpack-plugin`. I think that we are actually having a _slight_ bump in bundle size, as a result, but it's essentially negligible.

Also, `extract_css: true` is now required in order that `asset_pack_path('home_app.css')` actually returns a path to our CSS.

Side note... we seem to be building CSS assets twice, e.g. we have a file `home_app-1bafbc7d.css` and `home_app-5f66070482667de2461e.css`, both exactly the same size. This is probably wasting time during the deploy process? I'm not sure if this is new or something that has been happening for a while, unbeknownst to me. I think they must both be getting compiled by webpack, in that they both have e.g. `.modal-container[data-v-7b17699e]{transform:scale(1.1)}`.

### DIFF
--- a/config/webpack/production.js
+++ b/config/webpack/production.js
@@ -1,7 +1,6 @@
 const webpack = require('webpack');
 const merge = require('webpack-merge');
 const MiniCssExtractPlugin = require('mini-css-extract-plugin');
-const UglifyJsPlugin = require('uglifyjs-webpack-plugin');
 const OptimizeCSSAssetsPlugin = require('optimize-css-assets-webpack-plugin');
 const environment = require('./environment');
 const shared = require('./shared');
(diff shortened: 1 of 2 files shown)


config/webpack/production.js
```diff
@@ -23,11 +22,6 @@ const productionConfig = merge(environmentConfig, shared, {
   mode: 'production',
   optimization: {
     minimizer: [
-      new UglifyJsPlugin({
-        cache: true,
-        parallel: true,
-        sourceMap: true,
-      }),
       new OptimizeCSSAssetsPlugin({}),
```
davidrunger: quite possible this is what's somehow causing the duplication of CSS files being built... i should probably try removing this and seeing what happens...

PR #1227, opened 2019-01-04
Install specific version of bundler (1.17.3) on Travis

davidrunger:
Per the [release blog post][1], Heroku does not yet support bundler 2 (which was recently released), except with a custom buildpack. It seems safer to just stay on bundler 1 for now and wait for Heroku to add first class support for bundler 2 (which I am guessing will be relatively soon, anyway).

[1]: https://bundler.io/blog/2019/01/03/announcing-bundler-2.html


davidrunger:
For context, note that we used to install a specific version of bundler, but that was reverted in https://github.com/davidrunger/david_runger/pull/676/ .

PR #1281, opened 2019-01-26
Revert "Bump element-ui from 2.4.11 to 2.5.0"

davidrunger:
Reverts davidrunger/david_runger#1280

Something about the PR being reverted here broke the app:
image

See https://rollbar.com/davidjrunger/davidrunger/items/112/

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "clipboard": "^2.0.4",
     "compression-webpack-plugin": "^2.0.0",
     "css-loader": "^1.0.1",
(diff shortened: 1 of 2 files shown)


davidrunger:
It's surprising/confusing to me that somehow tests were still able to pass, despite this total failure -- wtf?

davidrunger:
Tests did fail; see https://github.com/davidrunger/david_runger/pull/1280#issuecomment-457927772 .

PR #1448, opened 2019-04-06
Refactor modal display logic and data structure

davidrunger:
This commit refactors the modal display logic and data structure, generalizing it to accommodate an arbitrary number of modals (as long as each modal is given a unique `name` prop).

This refactor also eliminates the necessity of adding a data attribute to indicate whether each specific modal is showing / not showing and eliminates the necessity of adding mutations for showing/hiding each modal that might be used.

There will be a slight performance penalty, in that we are now doing some array scanning (and other array operations) rather than manipulating specific booleans indicating the show/hide state of each individual modal, but this should be negligible. This performance penalty could be minimized somewhat by storing the show/hide state of each modal in an object rather than in an array, but then that would somewhat complicate the `hideTopModal` mutation, in contrast to the simplicity of just using an array as a stack where we know/assume that the last element in the array is the most recently opened modal.

This refactor is particularly motivated by the fact that we will be adding a third modal (which will contain a selector to choose a specific log to show, rather than the current system in which we show all logs at once).

### DIFF
--- a/spec/javascript/groceries/components/store.spec.js
+++ b/spec/javascript/groceries/components/store.spec.js
@@ -24,6 +24,7 @@ describe('Store', function () { // eslint-disable-line func-names, prefer-arrow-
       items: [{ id: 23, name: 'humus', needed: 2 }],
     };
     const bootstrap = {
+      current_user: { id: 2 },
       stores: [store],
     };
     wrapper = mount(
--- a/app/javascript/components/modal.vue
+++ b/app/javascript/components/modal.vue
@@ -1,5 +1,5 @@
 <template lang='pug'>
-transition(name='modal')
+transition(name='modal' v-if="showingModal({ modalName: name })")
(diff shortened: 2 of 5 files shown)


spec/javascript/groceries/components/store.spec.js
```diff
@@ -24,6 +24,7 @@ describe('Store', function () { // eslint-disable-line func-names, prefer-arrow-
       items: [{ id: 23, name: 'humus', needed: 2 }],
     };
     const bootstrap = {
+      current_user: { id: 2 },
```
davidrunger: this spec change was necessary because i guess that we are now rendering the `slot` content for all of the modals upfront, and the `need_phone_number_modal` component's `slot` requires `this.current_user.id` (for populating the `id` param in the `edit_user_path`).

from a performance perspective, that's not ideal to spend time calculating the content of the modal unnecessarily, but whatever, i guess. the upside is that we get the convenience of the modal knowing whether or not to render itself, based on its own `name`, by moving the `v-if` from the `Modal(name='foo-bar')` call into the `template` of the modal component itself.

PR #1450, opened 2019-04-06
Extract modal Vuex stuff into separate file

davidrunger:
This will allow us to "mix in" modal functionality into any Vuex store that we would like (e.g. the logs Vuex store, which currently doesn't have any modal-related functionality).

I considered using Vuex modules, but that seems to make things a lot more verbose and annoying. Thus, I'm just importing the modal Vuex functionality and splatting it into the groceries store. (I will do the same with the logs store in an upcoming PR.)

### DIFF
--- a/app/javascript/groceries/components/need_phone_number_modal.vue
+++ b/app/javascript/groceries/components/need_phone_number_modal.vue
@@ -17,7 +17,6 @@ export default {
   computed: {
     ...mapState([
       'current_user',
-      'showNeedPhoneNumberModal',
     ]),
 
     editUserPath() {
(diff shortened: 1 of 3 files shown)


app/javascript/groceries/components/need_phone_number_modal.vue
```diff
@@ -17,7 +17,6 @@ export default {
   computed: {
     ...mapState([
       'current_user',
-      'showNeedPhoneNumberModal',
```
davidrunger: unrelated cleanup that i am squeezing into this PR; this should have been removed in https://github.com/davidrunger/david_runger/pull/1448 (_Refactor modal display logic and data structure_)

PR #1451, opened 2019-04-06
Show only one log at a time; implement Slack-like log selector

davidrunger:
**See screencast:** http://screens.davidrunger.com/2019-04-06-14-17-32-0ewfl.mp4
This demonstrates using arrow keys to select a log, using fuzzy search, and clicking to select a log.

This change will:
1. lessen the memory/computational demands on the browser when a user has a lot of logs / log entries (by requiring the browser to only render one log at a time)
2. make it easier to navigate to a log using the fuzzy-search Slack-like selector implemented in this PR (as opposed to scrolling the page to find it, as before)

This also sets us up to not load all of the user's log entries into the `bootstrap` data (and instead request the log entries for a given log only when the log is viewed (or maybe eager-load the data via AJAX in the background)), which will provide a faster initial page load.

There's still a lot of work to be done, as well, in terms of improving the behavior after a user enters a new log entry. Currently, since we just reload `/logs`, the page reloads and the log that was being shown goes away; to see the new value in the log, the user would have to reselect the log.

I think that this PR lays a good foundation, though, in terms of setting us up to improve the above two issues/opportunities.

### DIFF
--- a/app/javascript/log/logs.vue
+++ b/app/javascript/log/logs.vue
@@ -2,25 +2,66 @@
 div
   .center.mb1
     .h5.gray.pt1 {{bootstrap.current_user.email}}
+    log-selector
     log(
-      v-for='log in bootstrap.logs'
-      :key='log.name'
-      :name='log.name'
-      :log_id='log.id'
-      :log_inputs='log.log_inputs'
-      :log_entries='log.log_entries'
+      v-if='selectedLog'
+      :key='selectedLog.name'
+      :name='selectedLog.name'
+      :log_id='selectedLog.id'
+      :log_inputs='selectedLog.log_inputs'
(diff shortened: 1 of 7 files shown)


app/javascript/log/logs.vue
```diff
@@ -2,25 +2,50 @@
...
+    div
+      a.js-link Create new log!
     log(
-      v-for='log in bootstrap.logs'
-      :key='log.name'
-      :name='log.name'
-      :log_id='log.id'
-      :log_inputs='log.log_inputs'
-      :log_entries='log.log_entries'
+      v-if='selectedLog'
```
davidrunger: i guess that here is really the key line that is being changed in this PR -- changing from `v-for='log in bootstrap.logs'` to `v-if='selectedLog'`

PR #1457, opened 2019-04-10
Add new log entries "live" via JavaScript rather than via page reload

davidrunger:
Much better! :D This change has been a long time coming.


davidrunger:
Weird; Travis wasn't running when I initially pushed this PR.

Deleting the branch (which necessitated first closing the PR) and then restoring it seems to have triggered a Travis build.

PR #1486, opened 2019-04-28
Lazy load log entries on-demand

davidrunger:
This allows us to remove the log entries from the bootstrap data. This should speed up the initial page load a lot and should overall be more efficient so that we don't spend time serializing log entries that might never be viewed.

### DIFF
--- a/.babelrc.js
+++ b/.babelrc.js
@@ -11,6 +11,7 @@ module.exports = {
   ],
 
   plugins: [
+    '@znck/prop-types/remove',
     '@babel/syntax-dynamic-import',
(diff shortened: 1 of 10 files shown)


davidrunger:
This has decreased the server time for `LogsController#index` from ~350ms to ~175ms -- so about twice as fast.

image

With the extra network time from the West Coast to the East Coast to fetch the log entries for each log ... this might not actually be super worth it? But maybe it is -- the initial page load speed is important. Additionally, if we proactively start eager-loading the log entries even before the user selects a log, that could help, too.

PR #1494, opened 2019-05-01
Speed up Travis builds

davidrunger:
Hopefully this will make Travis build times faster overall.


davidrunger:
I really don't think all of these changes have made much difference. Maybe a _small_ improvement on the order of 10 seconds or so.

I think that the main variable is just Travis. Sometimes it seems to run faster than other times. This also makes it hard to test whether my own changes are improving performance or not.

The best thing that can probably be done for spec performance at this point it to reduce our JavaScript dependencies to decrease the webpack compilation time.

PR #1505, opened 2019-05-08
Use Vue Router in logs app

davidrunger:
This will allow a user to navigate directly to the desired log by visiting `/logs/:log_slug` rather than having to go through the index page (`/logs`).

It also helps to keep the user "oriented" by reflecting the current log in the URL bar and by allowing navigation backwards and forwards using browser controls (including browser shortcut keys).


app/javascript/log/components/log.vue
```diff
@@ -73,16 +77,18 @@ export default {
...
+
+    ensureLogEntriesHaveBeenFetched() {
+      if (!this.log.log_entries) {
+        this.$store.dispatch('fetchLogEntries', { logId: this.log.id });
+      }
+    },
   },
 
-  props: {
-    log: PropTypes.shape({
```
davidrunger: removing these prop type validations is kind of a bummer.

i could create a sort of "wrapper" component that would extract the current log via the store via the URL, and then pass it to a child component as a prop, and the child component could then retain these prop validations, but that seems sort of like overkill.

it seems like maybe the thing to do would be if we could enforce some sort of validation in the store to check that the value returned by the `selectedLog` getter has the expected shape etc. Hopefully there is some library to facilitate that, but I haven't looked into it.

app/javascript/log/components/log_selector.vue
```diff
@@ -17,11 +17,17 @@ Modal(name='log-selector' width='85%', maxWidth='400px')
 </template>
 
 <script>
-import { mapState } from 'vuex';
+import { mapGetters, mapState } from 'vuex';
 import FuzzySet from 'fuzzyset.js';
 
+const ALL_LOGS = 'All logs';
```
davidrunger: this is a mostly unrelated feature addition, to allow going to the logs index via the quick-selector (which is accessed via a keyboard shortcut).

app/models/log.rb
```diff
@@ -4,13 +4,15 @@
 #
 #  created_at  :datetime         not null
 #  description :string
-#  id          :bigint(8)        not null, primary key
+#  id          :bigint           not null, primary key
```
davidrunger: unrelated changes introduced by `annotate` running (which was due to adding the `logs.slug` column)

spec/javascript/spec_helper.js
```diff
@@ -1,8 +1,5 @@
-import Vue from 'shared/customized_vue';
 import _ from 'lodash';
 
-window.Vue = Vue;
```
davidrunger: i guess this (and ` import Vue from 'shared/customized_vue';` above) wasn't necessary? tests are passing without this, and it seems like vue-test-utils encourages against importing a global vue and instead leaning on `createLocalVue` (which I think that I am doing in most (all?) specs already, anyway).

PR #1511, opened 2019-05-10
Fix (or ignore) all rubocop complaints; add rubocop check to Travis

davidrunger:


.travis.yml
```diff
@@ -25,6 +25,7 @@ script:
   - "yarn --version && [ \"$(yarn --version)\" == '1.15.2' ]"
   - yarn install
   - bin/rails spec:setup_js &
+  - bundle exec rubocop `git ls-tree -r HEAD --name-only` --force-exclusion
```
davidrunger: I am adding the `rubocop` at this point in the Travis build because it is after `bin/rails spec:setup_js` (so that process can continue to execute in the background during any CPU downtime for as long as possible) and it is otherwise as soon as possible in the build process to give early feedback on PRs.

.travis.yml
```diff
@@ -24,7 +24,8 @@ script:
   - "node --version && [ \"$(node --version)\" == 'v10.15.3' ]"
   - "yarn --version && [ \"$(yarn --version)\" == '1.15.2' ]"
   - yarn install
-  - bin/rails spec:setup_js &
+  - bin/rails spec:setup_js > /dev/null 2>&1 &
```
davidrunger: here i am quieting all output from this command, because otherwise it interferes with the rubocop output when there are errors

PR #1519, opened 2019-05-18
Focus log entry input whenever log/route is changed

davidrunger:
Previously, only the initial log that was viewed would have its input autofocused (which was done via the `mounted` lifecycle hook). Since vue router is now re-using the same component(s), the `mounted` hook was not being called when viewing any subsequent logs, and thus the log entry input was not being autofocused.

### DIFF
--- a/app/javascript/log/components/log.vue
+++ b/app/javascript/log/components/log.vue
@@ -86,7 +86,7 @@ export default {
   },
 
   watch: {
-    log() {
+    $route() {
       this.ensureLogEntriesHaveBeenFetched();
     },
   },
--- a/app/javascript/log/components/new_log_entry_form.vue
+++ b/app/javascript/log/components/new_log_entry_form.vue
@@ -29,6 +29,12 @@ export default {
   },
 
   methods: {
+    focusLogEntryInput() {
+      setTimeout(() => {
+        this.$refs['log-input'][0].focus();
+      });
+    },
+
     postNewLogEntry() {
       if (this.formstate.$invalid) return;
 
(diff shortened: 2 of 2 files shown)


app/javascript/log/components/log.vue
```diff
@@ -86,7 +86,7 @@ export default {
   },
 
   watch: {
-    log() {
+    $route() {
```
davidrunger: this is a mostly unrelated change. i think it's better to watch `$route` than to watch `log` because
1. it's a little clearer about why the log would be changing (because the route has changed)
2. it doesn't require that we have the `log` at all; all components have `$route`. it doesn't matter as much in this particular case, since the `Log` component does need/have a `log` anyway, but this argument makes more sense in the `NewLogEntryForm` component, which doesn't have/need a reference to the `log`. we could get it from the vuex store, but that is unnecessary when we can just use the `$route` instead. so using `$route` allows us to have a consistent pattern throughout the app while avoiding unnecessarily pulling data from the vuex store just to handle route changes that re-use/transform already rendered/mounted components.

i guess that a downside is that if the component's log changes for some _other_ reason besides the route changing, then we will no longer call `nesureLogEntriesHaveBeenFetched`, but since that is not expected to happen, I am not too worried about it.

app/javascript/log/components/new_log_entry_form.vue
```diff
@@ -44,9 +50,7 @@ export default {
   },
 
   mounted() {
-    setTimeout(() => {
-      this.$refs['log-input'][0].focus();
-    });
+    this.focusLogEntryInput();
```
davidrunger: i could use `autofocus` in the template for this (instead of `mounted`), but it seems better to stay consistent and always do the focusing via JavaScript.

PR #1520, opened 2019-05-18
Fix undefined log name in delete-last-log-entry confirmation message

davidrunger:
# Before

image

# After

image

### DIFF
--- a/app/javascript/log/components/log.vue
+++ b/app/javascript/log/components/log.vue
@@ -68,7 +68,7 @@ export default {
   methods: {
     destroyLastEntry() {
       var confirmation = confirm(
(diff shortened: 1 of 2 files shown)


davidrunger:
This bug was introduced via this change, I think: https://github.com/davidrunger/david_runger/commit/e522ab38d9935278cc1f5db7e07c922a6da5d538#diff-b8e0c29abca6d5ed1f3add9e5c9a0accR79

That change in https://github.com/davidrunger/david_runger/pull/1486 (_Lazy load log entries on-demand_) started just passing a `log` prop to the `Log` component, rather than passing each prop individually (`log_name`, `log_id`, etc.).

Tests should have protected against this bug, but they were lacking. A test has been added in this PR.

PR #1524, opened 2019-05-20
Add back Log#log_entries etc associations needed for datamigration

davidrunger:
This fixes an issue with one of the datamigrations added in https://github.com/davidrunger/david_runger/pull/1523 .

We will need to clean this up later.

### DIFF
--- a/app/models/log.rb
+++ b/app/models/log.rb
@@ -40,6 +40,15 @@ class Log < ApplicationRecord
 
   belongs_to :user
 
+  has_many :log_entries,
+    dependent: :destroy,
+    inverse_of: :log
+  has_many :log_entries_ordered,
+    -> { order(:created_at) },
+    class_name: 'LogEntry',
+    dependent: :destroy,
+    inverse_of: :log
+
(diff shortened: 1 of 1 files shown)


app/models/log.rb
```diff
@@ -67,11 +76,19 @@ class Log < ApplicationRecord
   before_save :set_slug, if: -> { name_changed? }
 
   def log_entries
-    public_send(DATA_TYPES[data_type][:association])
+    if respond_to?(:data_type)
```
davidrunger: (this will basically depend on whether the migration adding this column has yet been executed)

app/models/log.rb
```diff
@@ -67,11 +76,19 @@ class Log < ApplicationRecord
...
+      public_send(DATA_TYPES[data_type][:association])
+    else
+      super
+    end
   end
 
   def log_entries_ordered
-    public_send(DATA_TYPES[data_type][:ordered_association])
+    if respond_to?(:data_type)
```
davidrunger: (same as above)

PR #1526, opened 2019-05-20
Run eslint earlier in Travis build process

davidrunger:
I don't know why I originally added it so late in the build process?


davidrunger:
Oh, I see, it's because we need to wait for `webpack.config.static.js` to be built, so that we can resolve imports correctly.

Closing.

PR #1542, opened 2019-05-26
Force Vue router to render fresh components via `key`

davidrunger:
The fact that components were being re-used was requiring `watch`ing the `$route` for changes (to then `ensureLogEntriesHaveBeenFetched` and `focusLogEntryInput`). It was also causing a bug with our `vue-form` form validations; for example, after viewing a number-type log, then switching to a text log, `vue-form` was validating the text log's log entry input value as a number (i.e. still left over from the number log input validation).

### DIFF
--- a/app/javascript/logs/components/log.vue
+++ b/app/javascript/logs/components/log.vue
@@ -82,12 +82,6 @@ export default {
       }
     },
   },
-
-  watch: {
-    $route() {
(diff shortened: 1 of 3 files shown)


davidrunger:
I don't _think_ that this will have too big of a performance penalty, since I don't really think that probably too much of the component(s) were getting re-used when switching from log to log?

PR #1547, opened 2019-05-30
Add frozen_string_literal to all linted files; enable rubocop cop

davidrunger:
This should reduce the need for garbage collection and thereby increase performance.

Good summary: https://www.mikeperham.com/2018/02/28/ruby-optimization-with-one-magic-comment/

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -60,7 +60,7 @@ Style/DocumentationMethod:
 Style/EmptyCaseCondition:
   Enabled: false
 Style/FrozenStringLiteralComment:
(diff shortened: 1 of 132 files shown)


davidrunger:
I don't know why I originally disabled this rule in https://github.com/davidrunger/david_runger/pull/17 (_Add rubocop, tweak some rules, conform the Gemfile_). Probably some combination of not thinking about rubocop's autocorrect (and not wanting to manually fix all of the files) and not considering / knowing about the potential benefits in terms of memory savings / performance.

PR #1557, opened 2019-05-31
Set up feature tests; add a basic feature test for home page

davidrunger:
Although this PR doesn't look like much, it wasn't easy to get everything working just right!, but I will definitely feel more confident having feature tests and look forward to writing more feature tests in the future to cover important flows and/or flows where bugs emerge (or are discovered) in the future.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ rvm:
   - "2.6.3"
 services:
   - postgresql
+  - redis-server
 addons:
   postgresql: "9.6"
 env:
@@ -18,6 +19,9 @@ before_install:
   - nvm install 10.15.3
   - curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.15.2
   - export PATH="$HOME/.yarn/bin:$PATH"
+  - sudo apt-get install chromium-chromedriver
+before_script:
+  - export PATH=$PATH:/usr/lib/chromium-browser/
 script:
   - ruby --version && [ "$(ruby --version | cut -c1-11)" == 'ruby 2.6.3p' ]
   - bundle --version && [ "$(bundle --version)" == 'Bundler version 1.17.3' ]
@@ -30,10 +34,10 @@ script:
   - bin/rails db:create
   - bin/rails db:schema:load
   - bin/rails spec:fixture_builder:rebuild
-  - bin/rails spec:rb
   - bin/rails spec:poll_js
-  - bin/rails spec:run_js
   - ./node_modules/.bin/eslint --max-warnings 0 --ext .js,.vue app/javascript/ spec/javascript/
+  - bin/rails spec:run_js
+  - bin/rails spec:rb
 notifications:
   email:
     on_success: never
--- a/lib/tasks/spec.rake
+++ b/lib/tasks/spec.rake
@@ -13,7 +13,11 @@ end
 namespace :spec do
   desc 'Run Ruby specs'
   task :rb do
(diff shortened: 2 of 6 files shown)


.travis.yml
```diff
@@ -30,10 +34,10 @@ script:
   - bin/rails db:create
   - bin/rails db:schema:load
   - bin/rails spec:fixture_builder:rebuild
-  - bin/rails spec:rb
   - bin/rails spec:poll_js
-  - bin/rails spec:run_js
   - ./node_modules/.bin/eslint --max-warnings 0 --ext .js,.vue app/javascript/ spec/javascript/
+  - bin/rails spec:run_js
+  - bin/rails spec:rb
```
davidrunger: i've switched the ordering here (`spec:run_js` now runs before `spec:rb`) because these two things now conflict with each other by writing different files into `public/packs-test/`. so basically we can't have `spec:rb` running in between `bin/rails spec:setup_js` (above on L31) and `spec:run_js`. that's why i moved `spec:run_js` above `spec:rb`.

lib/tasks/spec.rake
```diff
@@ -58,7 +62,13 @@ namespace :spec do
 
   desc 'Run JavaScript tests (after setting up already)'
   task :run_js do
+    # run the tests
     run_logged_system_command('yarn run test')
+
+    # kill JS unit test server (if running)
+    run_logged_system_command(
+      "ps -ax | egrep 'ruby.*httpd' | egrep -v egrep | awk '{print $1}' | xargs kill",
```
davidrunger: `ps` options are not very compatible between Linux and Mac, it seems. (At first i was using `ps -e` on my Mac, but that wasn't finding the process on Travis (Linus)). `ps -ax` seems to produce the desired result on both systems. 👍

lib/tasks/spec.rake
```diff
@@ -13,7 +13,11 @@ end
 namespace :spec do
   desc 'Run Ruby specs'
   task :rb do
+    Rake::Task['spec:copy_production_webpacker_settings_to_test'].
+      invoke('cache_manifest compile extract_css source_path')
```
davidrunger: this part is nasty ... basically we are using webpacker's `test` environment for two different purposes:
1. compile JavaScript unit test framework
2. compile the app for feature tests

and we need different settings in `webpacker.yml` for each of those purposes. `webpacker.yml` doesn't allow ERB, though, so we can't use that to adjust the settings. so we are literally just changing the file contents temporarily here.

it would almost work to just use the `production` webpacker settings to compile the app (by temporarily settings `RAILS_ENV=production`) but i don't think that would work because that would compile the assets to `public/packs/`, i think, but we need them to be in `public/packs-test/` so that our feature test(s) can find them there. I guess i could just copy the assets from `public/packs/` to `public/packs-test/`, though ... However, another reason to not use the production config here is so as not to spend/waste additional time minifying the JS and CSS assets.

lib/tasks/spec.rake
```diff
@@ -31,7 +35,7 @@ namespace :spec do
 
   desc 'Poll until JavaScript test server bootup is confirmed'
   task :poll_js do
-    max_polls = 201
+    max_polls = 401
```
davidrunger: i'm increasing the poll time to 80 seconds because there's no longer as much time for the compilation to happen before we start polling, since i had to move `spec:run_js` higher up in the build process (reason for that described here: https://github.com/davidrunger/david_runger/pull/1557/files#r289591592).

PR #1567, opened 2019-06-01
Remove ApplicationJob

davidrunger:
The `ApplicationJob` definition references `ActiveJob::Base` which is no longer being loaded as of c51a40b / #1566. It's causing the application to fail to boot in production. This only shows up in production because of eager loading; in development we simply weren't trying to load `ApplicationJob`.

Since we aren't using `ApplicationJob`, the fix is just to remove it.

### DIFF
--- a/app/jobs/application_job.rb
+++ b/app/jobs/application_job.rb
@@ -1,4 +1,0 @@
-# frozen_string_literal: true
-
-class ApplicationJob < ActiveJob::Base
-end


davidrunger:
> It's causing the application to fail to boot in production.

This seems to highlight some inadequate test coverage. Maybe we aren't eager-loading in `test`? Maybe we should?

davidrunger:
> this seems to highlight some inadequate test coverage

Fixed by https://github.com/davidrunger/david_runger/pull/1569/ (_Eager load the app in test on Travis CI_).

PR #1568, opened 2019-06-01
Remove ApplicationMailer

davidrunger:
This references `ActionMailer::Base` which is no longer being loaded as of c51a40b / #1566. It's causing the application to fail to boot in production.

### DIFF
--- a/app/mailers/application_mailer.rb
+++ b/app/mailers/application_mailer.rb
@@ -1,6 +1,0 @@
-# frozen_string_literal: true
-
-class ApplicationMailer < ActionMailer::Base
-  default from: 'from@example.com'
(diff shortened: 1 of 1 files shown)


davidrunger:
This is basically the same issue as https://github.com/davidrunger/david_runger/pull/1567/ (_Remove ApplicationJob_).

As I also commented on that PR: this seems to highlight some inadequate test coverage. Maybe we aren't eager-loading in test? Maybe we should?

davidrunger:
> this seems to highlight some inadequate test coverage

Fixed by https://github.com/davidrunger/david_runger/pull/1569/ (_Eager load the app in test on Travis CI_).

PR #1569, opened 2019-06-01
Eager load the app in test on Travis CI

davidrunger:
This should catch production-breaking bugs like the one(s) that were caused by #1566 / c51a40b and fixed by #1567 / a0c8a2a and #1568 / fe2e33c .


davidrunger:
Unfortunately, I think that this might add a little bit of test execution time. That's okay -- worth it!

PR #1570, opened 2019-06-01
Render correct template (index, not show) for groceries#index action

davidrunger:
This fixes a 500 error that is occurring when trying to view the groceries app.

### DIFF
--- a/lib/tasks/spec.rake
+++ b/lib/tasks/spec.rake
@@ -13,11 +13,14 @@ end
 namespace :spec do
   desc 'Run Ruby specs'
   task :rb do
-    Rake::Task['spec:copy_production_webpacker_settings_to_test'].
-      invoke('cache_manifest compile extract_css source_path')
-    run_logged_system_command('bin/webpack --silent')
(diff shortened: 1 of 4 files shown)


lib/tasks/spec.rake
```diff
@@ -13,11 +13,14 @@ end
...
-    run_logged_system_command('bin/webpack --silent')
-    run_logged_system_command('bin/rspec --format documentation')
-    run_logged_system_command('git checkout config/webpacker.yml')
+    begin
+      Rake::Task['spec:copy_production_webpacker_settings_to_test'].
+        invoke('cache_manifest compile extract_css source_path')
+      run_logged_system_command('RAILS_ENV=test NODE_ENV=test bin/webpack --silent')
+      run_logged_system_command('bin/rspec --format documentation')
+    ensure
+      run_logged_system_command('git checkout config/webpacker.yml')
```
davidrunger: we want to restore `config/webpacker.yml` even if a test (or other step above) fails

PR #1572, opened 2019-06-01
Upgrade to Rails 6.0.0.rc1

davidrunger:
That wasn't all too hard! :)


Gemfile
```diff
@@ -26,14 +30,13 @@ gem 'pg_query', '>= 0.9.0'
...
 gem 'pundit'
-gem 'rails', '>= 5.2.0.rc1'
+gem 'rails', '>= 6.0.0.rc1'
 gem 'redis', '~>4.1'
 gem 'rest-client'
 gem 'rollbar'
 gem 'rubocop', require: false
 gem 'rubocop-performance', require: false
 gem 'rubocop-rails', require: false
-gem 'sass-rails', '~> 5.0'
```
davidrunger: this gem (`sass-rails`) depends on `sass`, which was giving this post-install warning:
```
Post-install message from sass:

Ruby Sass has reached end-of-life and should no longer be used.

* If you use Sass as a command-line tool, we recommend using Dart Sass, the new
  primary implementation: https://sass-lang.com/install

* If you use Sass as a plug-in for a Ruby web framework, we recommend using the
  sassc gem: https://github.com/sass/sassc-ruby#readme

* For more details, please refer to the Sass blog:
  https://sass-lang.com/blog/posts/7828841
```

Since I am not using sprockets / the asset pipeline at all anymore, I figured that maybe I don't even need it?, and indeed everything seems to work without it.

app/controllers/api/log_entries_controller.rb
```diff
@@ -6,8 +6,6 @@ def create
     log_entry_params = params.require(:log_entry).permit(:log_id, :data)
     @log_entry = log.log_entries.build(log_entry_params)
     if @log_entry.save
-      # TODO: remove #reload; needed to work w/ #read_attribute_before_type_cast call in serializer
-      @log_entry.reload
```
davidrunger: this reload is no longer needed because of the change in `app/serializers/log_entry_serializer.rb`

config/application.rb
```diff
@@ -40,6 +40,7 @@ class DavidRunger::Application < Rails::Application
   ]
   config.autoload_paths.concat(extra_load_paths)
   config.eager_load_paths.concat(extra_load_paths)
+  config.add_autoload_paths_to_load_path = false
```
davidrunger: See here: https://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#config-add-autoload-paths-to-load-path

> The new configuration point `config.add_autoload_paths_to_load_path` is true by default for backwards compatibility, but allows you to opt-out from adding the autoload paths to $LOAD_PATH.
> 
> This makes sense in most applications, since you never should require a file in app/models, for example, and Zeitwerk only uses absolute file names internally.
> 
> By opting-out you optimize $LOAD_PATH lookups (less directories to check), and save Bootsnap work and memory consumption, since it does not need to build an index for these directories.

config/initializers/assets.rb
```diff
@@ -10,8 +10,6 @@
 # Add Yarn node_modules folder to the asset load path.
 Rails.application.config.assets.paths << Rails.root.join('node_modules')
 
-Rails.application.config.assets.precompile += %w[routes.js]
```
davidrunger: I don't know why I added this; it doesn't seem to be necessary.

PR #1574, opened 2019-06-02
Make textarea input for text logs a little taller to start with

davidrunger:
Mostly this is because I want to be able to enter a list of 3 items (separated by newlines) for my gratitude journal.


app/javascript/logs/components/new_log_entry_form.vue
```diff
@@ -72,8 +72,10 @@ export default {
 };
 </script>
 
-<style>
-textarea {
-  font-family: sans-serif;
+<style lang='scss' scoped>
+/deep/ .new-log-input {
```
davidrunger: the `/deep/` is required here because technically the `textarea` is being rendered within a different component, the `el-input` element-ui component

PR #1578, opened 2019-06-02
Give modal a default class of `bg-white`, not `.bg-white`

davidrunger:
This was causing the modal to be transparent in the groceries app.

### DIFF
--- a/app/javascript/components/modal.vue
+++ b/app/javascript/components/modal.vue
@@ -6,7 +6,7 @@ transition(name='modal' v-if="showingModal({ modalName: name })")
   )
     div.modal-container.p3.rounded(
       :style='{ width: width, maxWidth: maxWidth }'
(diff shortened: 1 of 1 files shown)


app/javascript/components/modal.vue
```diff
@@ -6,7 +6,7 @@ transition(name='modal' v-if="showingModal({ modalName: name })")
   )
     div.modal-container.p3.rounded(
       :style='{ width: width, maxWidth: maxWidth }'
-      :class='this.backgroundClass || ".bg-white"'
```
davidrunger: i moved this to a default prop, rather than this `||` here. that's just a refactoring. the real change is `.bg-white` => `bg-white`.

davidrunger:
Travis is having issues. Since this is basically just a visual change, I am going to merge.

PR #1579, opened 2019-06-02
Switch to new `ActiveSupport::ParameterFilter` class

davidrunger:
(...from `ActionDispatch::Http::ParameterFilter`).

This is deprecated; see message that is logged 7 times when running `bin/rails spec:rb`:
> DEPRECATION WARNING: ActionDispatch::Http::ParameterFilter is deprecated and will be removed from Rails 6.1. Use ActiveSupport::ParameterFilter instead. (called from filtered_params at app/controllers/application_controller.rb:59)

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -56,7 +56,7 @@ class ApplicationController < ActionController::Base
   end
 
   def filtered_params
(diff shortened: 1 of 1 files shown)


davidrunger:
Travis still seems to be having issues; I am going to go ahead and merge this. I ran the `spec:rb` tests locally.

PR #1580, opened 2019-06-02
Use Rails 6 defaults

davidrunger:
Note that this will also switch us to the `zeitwerk` loader.

> The default configuration for Rails 6
> ```
> # config/application.rb
> config.load_defaults "6.0"
> ```
> 
> enables `zeitwerk` autoloading mode on CRuby. In that mode, autoloading, reloading, and eager loading are managed by [Zeitwerk](https://github.com/fxn/zeitwerk).

-- https://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-5-2-to-rails-6-0


config/initializers/new_framework_defaults_6_0.rb
```diff
@@ -1,37 +0,0 @@
-# frozen_string_literal: true
```
davidrunger: As I understand it, the purpose of this file (which was generated via `bin/rails app:update` when migrating from 5.2 to 6.0) is to assist with gradually switching the configuration options below one at a time. Most of the configuration options below, however, are not applicable to our application. Therefore, they all "just work" (or, more typically, I think, simply have no effect). So I've simply changed `config.load_defaults(5.1)` to `config.load_defaults('6.0')`, and thus this file is no longer needed.

davidrunger:
It seems that the Travis tests are failing because of a loading issue (which I am guessing is related to the fact that we are now using the `zeitwork` loader).

davidrunger:
> I can't seem to reproduce the issue when running bin/rails spec:rb locally

I can reproduce locally by setting `CI=true bin/rails spec:rb`; in other words, I think that one or both of these settings are causing this issue on CI (and probably will also cause this issue when trying to boot in `production`): https://github.com/davidrunger/david_runger/blob/9bdc854d52ad932d835621d1a2401b684f9d5da4/config/environments/test.rb#L11-L16

davidrunger:
I found a solution to the Zeitwerk eager-loading issue; see here: https://github.com/davidrunger/david_runger/pull/1580/files#diff-b1fe55db50c712fef0673345e5b9c0d9R48

.rubocop.yml
```diff
@@ -40,6 +40,10 @@ Metrics/AbcSize:
 Metrics/BlockLength:
   Enabled: false
 Metrics/LineLength:
+  IgnoredPatterns:
+    # ignore line length if the line is a comment without any spaces; it's probably not something we
+    # can fix (e.g. a long file path)
+    - !ruby/regexp /^ *#? [\S]+$/
```
davidrunger: I'm adding this `IgnoredPattern` so that rubocop doesn't complain about the comment line being added in this PR that includes the lengthy file path `/home/travis/build/davidrunger/david_runger/vendor/bundle/ruby/2.6.0/gems/devise-4.6.2/app/mailers/devise/mailer.rb` (and so that it doesn't complain about similar situations in the future).

PR #1581, opened 2019-06-02
Tweak some copy and other things in the 500 error page

davidrunger:
# Before

image

# After

image

# Also

Not seen in the screenshots is the fact that the `title` of the page has been changed from `We're sorry, but something went wrong (500)` to `500 Error - davidrunger.com`.

### DIFF
--- a/public/500.html
+++ b/public/500.html
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html>
 <head>
-  <title>We're sorry, but something went wrong (500)</title>
+  <title>500 Error - davidrunger.com</title>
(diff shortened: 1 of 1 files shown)


public/500.html
```diff
@@ -13,7 +13,7 @@
 
   .rails-default-error-page div.dialog {
     width: 95%;
-    max-width: 33em;
+    max-width: 36em;
```
davidrunger: I increased this `max-width` a little bit so that the `Sorry, something went wrong.` `<h1>` text can all fit on a single line.

PR #1606, opened 2019-06-08
Enforce that there is no space before JavaScript function args paren

davidrunger:
... via a new `'space-before-function-paren': ['warn', 'never']` eslint rule.

... and auto-fix existing violations:
```
./node_modules/.bin/eslint  --fix --max-warnings 0 --ext .js,.vue app/javascript/ spec/javascript/
```

### DIFF
--- a/app/javascript/logs/components/data_renderers/duration_timeseries.vue
+++ b/app/javascript/logs/components/data_renderers/duration_timeseries.vue
@@ -70,13 +70,15 @@ export default {
         }],
         yAxes: [{
           ticks: {
-            userCallback: function(v) { return epochMsToHhMmSs(v); },
+            userCallback(v) {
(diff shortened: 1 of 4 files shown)


app/javascript/logs/components/data_renderers/duration_timeseries.vue
```diff
@@ -70,13 +70,15 @@ export default {
         }],
         yAxes: [{
           ticks: {
-            userCallback: function(v) { return epochMsToHhMmSs(v); },
```
davidrunger: this function (and the `label` one below) were brought to my attention because they violate the new configuration for the `space-before-function-paren` rule; i fixed by changing to the method/function-in-object syntax (whatever it's formally called...).

PR #1619, opened 2019-06-13
Manually install specific chromedriver version on Travis

davidrunger:


lib/tasks/spec.rake
```diff
@@ -73,7 +73,7 @@ namespace :spec do
 
     # kill JS unit test server (if running)
     run_logged_system_command(
-      "ps -ax | egrep 'ruby.*httpd' | egrep -v egrep | awk '{print $1}' | xargs kill",
+      "ps -ax | egrep 'ruby.*httpd' | egrep -v grep | awk '{print $1}' | xargs kill",
```
davidrunger: unrelated change ... not sure why this is needed now. it seems that the `egrep 'ruby.*httpd'` was creating a `grep -E ruby.*httpd` command that then wasn't being filtered by the `egrep -v egrep` (since the command is `grep`, not `egrep`). as a result, we then tried to pipe two PIDs to `xargs kill` -- one for the `grep` and one for the `ruby -run -ehttpd public -p8080` process. but then when the `kill`ing happened, only the `ruby` process was still running, so the `kill`ing of the no-longer-existent `grep` command failed / gave an exit code of `1`. something like that...

.travis.yml
```diff
@@ -19,9 +20,9 @@ before_install:
   - nvm install 10.15.3
   - curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.15.2
   - export PATH="$HOME/.yarn/bin:$PATH"
-  - sudo apt-get install chromium-chromedriver
-before_script:
-  - export PATH=$PATH:/usr/lib/chromium-browser/
+  # To get a different version of chromedriver, see http://chromedriver.chromium.org/downloads
+  - curl -o tmp/chromedriver.zip https://chromedriver.storage.googleapis.com/75.0.3770.8/chromedriver_linux64.zip
```
davidrunger: i don't love this ... it seems like i'll need to update this chromedriver URL whenever Chrome releases a new version. i couldn't find a generic "stable" URL for `chromedriver` from Google, though. I think that there are various packages that try to basically provide this (both in `npm` and in Ruby-land), but then i feel like i'll just be delegating this problem to whichever of those packages that i might use; i'd rather at least have control myself. so whatever, let's try this for now and see how it goes / how annoying it is to update this URL whenever Chrome releases a new version.

.travis.yml
```diff
@@ -10,6 +10,7 @@ services:
   - postgresql
   - redis-server
 addons:
+  chrome: stable
```
davidrunger: without this, Travis was running version 73, it seems, which caused issues because `sudo apt-get install chromium-chromedriver` has started installing a chromedriver version that is only compatible with Chrome 74.

this change means that we now are running Chrome 75, and i've switched (see below) to manually installing a chromedriver version that also works with Chrome 75.

.travis.yml
```diff
@@ -19,9 +20,9 @@ before_install:
   - nvm install 10.15.3
   - curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.15.2
   - export PATH="$HOME/.yarn/bin:$PATH"
-  - sudo apt-get install chromium-chromedriver
```
davidrunger: see this comment https://github.com/davidrunger/david_runger/pull/1619/files#r293209096 for an explanation of why this was problematic and how this PR fixes it.

.travis.yml
davidrunger: for evidence/details see how it says `Unpacking google-chrome-stable (75.0.3770.80-1) over (73.0.3683.86-1) ...` here https://travis-ci.org/davidrunger/david_runger/builds/545065123 (which happens now that we have this `chrome: stable` addon); i think this shows that the version previously installed by default was version 73.

PR #1627, opened 2019-06-17
Add notes for duration logs and number logs

davidrunger:


Gemfile
```diff
@@ -68,9 +68,12 @@ end
 
 group :test do
   gem 'capybara'
+  # note: to use guard-espect from command line, it will also have to be installed "globally"
   gem 'guard-espect', require: false, github: 'davidrunger/guard-espect'
   # TEMP: list hashdiff explicitly in Gemfile to force 1.0.0.beta1; remove once 1.0.0 is released
   gem 'hashdiff', '1.0.0.beta1'
+  # for testing ActiveModelSerializers
+  gem 'json-schema'
```
davidrunger: in addition to using `json-schema` to test ActiveModelSerializers, i am also thinking that I can use this to define/enforce a contract between the server and the client for unit tests. in other words, when a javascript test stubs a server response, make sure that the stubbed server response conforms to the json schema. and then in controller tests, ensure that the responses conform to the json schema.

app/controllers/api/log_entries_controller.rb
```diff
@@ -3,12 +3,12 @@
...
   def create
     log = current_user.logs.find(params.dig(:log_entry, :log_id))
-    log_entry_params = params.require(:log_entry).permit(:log_id, :data)
+    log_entry_params = params.require(:log_entry).permit(:data, :log_id, :note)
     @log_entry = log.log_entries.build(log_entry_params)
     if @log_entry.save
-      render json: @log_entry
+      render json: @log_entry, status: :created
     else
-      render json: {errors: @log_entry.errors.to_h}, status: :unprecessable_entity
```
davidrunger: typo caught by adding some missing test coverage here!

PR #1748, opened 2019-08-04
Prevent decrementing needed number of items to a negative value

davidrunger:


app/javascript/groceries/components/item.vue
```diff
@@ -34,6 +34,13 @@ export default {
   },
 
   methods: {
+    decrement(item) {
+      const newNeededCount = item.needed - 1;
+      if (newNeededCount >= 0) {
```
davidrunger: the reason that I put this check in a new `decrement` method, rather than putting it in the existing `setNeeded` method, is because users might have already decremented some of their needed counts of grocery items to a negative number. we need to allow those users to change (increase) those numbers from e.g. -3 to -2, which would have been prevented if we had put this check in `setNeeded` (which is used both when incrementing and when decrementing).

PR #1763, opened 2019-08-11
Remove sourceMap option from style-loader config in development

davidrunger:
This should fix:
> ValidationError: Invalid options object. Style Loader has been
> initialised using an options object that does not match the API
> schema. [...] options has an unknown property 'sourceMap'.

Per https://github.com/webpack-contrib/style-loader/releases/tag/v1.0.0
> the `sourceMap` option was removed. The loader automatically inject
> source maps if the previous loader emit them

### DIFF
--- a/config/webpack/development.js
+++ b/config/webpack/development.js
@@ -11,9 +11,6 @@ environment.loaders.append('style', {
   use: [
     {
       loader: 'style-loader',
-      options: {
-        sourceMap: true,
-      },
     },
     {
       loader: 'css-loader',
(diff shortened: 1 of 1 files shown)


config/webpack/development.js
```diff
@@ -1,4 +1,3 @@
-const webpack = require('webpack');
```
davidrunger: (unrelated cleanup)

davidrunger:
This fixes an error that causes webpack compilation in development to fail. This error was introduced in https://github.com/davidrunger/david_runger/pull/1754 (_Bump style-loader from 0.23.1 to 1.0.0_).

PR #1766, opened 2019-08-11
Close logs quick-selector when route changes; stop managing manually

davidrunger:
This change fixes a bug wherein command-clicking (Mac) a log link from within the logs quick-selector would close the quick-selector. The quick selector should remain open so that multiple logs can be opened from the quick selector via command-clicking.

Since command-clicking doesn't trigger a route change within the current tab, this change (close the logs quick-selector when the route changes, and stop managing the appearance of the quick-selector manually) results in the quick selector remaining open when command-clicking links, as desired.

At first I was going to move the quick-selector state (`highlightedLogIndex`, `searchString`) into the Vuex store, so that the router could dispatch Vuex actions to reset the quick selector state when the route changes.

However, I ended up instead just having the router emit a `'groceries:route-changed'` event that can be listened for by the quick selector, and then the quick selector can manage resetting its own state. This was a simpler/smaller change, and I think it largely "feels right" in terms of the router shouldn't necessarily "know" about Vuex actions to update the state of the logs quick-selector.

### DIFF
--- a/app/javascript/groceries/store.js
+++ b/app/javascript/groceries/store.js
@@ -76,6 +76,8 @@ const getters = {
   ...ModalVuex.getters,
 
   currentStore(state) {
+    if (!state.stores) return null;
+
     return (
       _(state.stores).filter('viewed_at').sortBy(['viewed_at']).last() ||
       state.stores[0]
--- a/app/javascript/groceries/store.js
+++ b/app/javascript/groceries/store.js
@@ -111,3 +111,5 @@ export function groceryVuexStoreFactory(bootstrap) {
     mutations,
   });
 }
+
+export default groceryVuexStoreFactory(window.davidrunger.bootstrap);
--- a/app/javascript/logs/components/log_selector.vue
+++ b/app/javascript/logs/components/log_selector.vue
@@ -19,13 +19,14 @@ Modal(
         router-link.log-link(
           :to='{ name: "log", params: { slug: log.slug }}'
           :class='{bold: (index === highlightedLogIndex)}'
-          @click.native='resetQuickSelector'
         ) {{log.name}}
 </template>
 
 <script>
 import { mapGetters, mapState } from 'vuex';
 import FuzzySet from 'fuzzyset.js';
+
+import { on } from 'lib/event_bus';
 
 export default {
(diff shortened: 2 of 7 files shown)


app/javascript/groceries/store.js
```diff
@@ -76,6 +76,8 @@ const getters = {
   ...ModalVuex.getters,
 
   currentStore(state) {
+    if (!state.stores) return null;
```
davidrunger: (mostly?) unrelated cleanup. i was seeing something about "can't read property 0 of undefined" in the JavaScript console. it didn't seem to have a negative impact, though, i.e. presumably it was only occurring before the stores got loaded into the groceries Vuex store.

app/javascript/groceries/store.js
```diff
@@ -109,3 +111,5 @@ export function groceryVuexStoreFactory(bootstrap) {
     mutations,
   });
 }
+
+export default groceryVuexStoreFactory(window.davidrunger.bootstrap);
```
davidrunger: basically unrelated change. allows referencing the groceries vuex store from multiple different files, to have the canonical store be actually created in one place, here.

app/javascript/logs/components/log_selector.vue
```diff
@@ -19,14 +19,15 @@ Modal(
         router-link.log-link(
           :to='{ name: "log", params: { slug: log.slug }}'
           :class='{bold: (index === highlightedLogIndex)}'
-          @click.native='resetQuickSelector'
```
davidrunger: basically it's the removal of this line that allows the user to CMD-click log links from quick-selector without closing the modal on click.

spec/javascript/mocha_runner.html
```diff
@@ -7,14 +7,14 @@
 <body>
   <div id="mocha"></div>
   <script>
-    window.davidrunger = { env: 'test' };
+    window.davidrunger = { env: 'test', bootstrap: {} };
```
davidrunger: this was necessary to fix failing javascript unit tests and failing feature tests, very likely because of the change to `app/javascript/groceries/store.js` in which we now have a default export which is a vuex store instance.

PR #1804, opened 2019-08-31
Remove foreman from Gemfile

davidrunger:
Instead, we will install foreman separately from the bundled gems using the [heroku-buildpack-run][1] buildpack, which I have added to the Heroku app via `heroku buildpacks:add https://github.com/weibeld/heroku-buildpack-run`. That buildpack will execute `buildpack-run.sh`, which is also added in this commit.

Also, update to Ruby 2.7.0-preview1, because I wanted it for `Module#const_source_location` to debug this change.

[1]: https://elements.heroku.com/buildpacks/weibeld/heroku-buildpack-run

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,2 +1,2 @@
-web: bundle exec foreman start -f Procfile.multi
+web: vendor/heroku/ruby/2.6.0/gems/foreman-0.85.0/bin/foreman start -f Procfile.multi
 release: bin/release-tasks
--- a/buildpack-run.sh
+++ b/buildpack-run.sh
@@ -1,0 +1,9 @@
+#!/usr/bin/env bash
+
+gem_parent_dir="vendor/heroku/ruby/2.6.0"
+
+mkdir -p $gem_parent_dir
+mkdir -p vendor/heroku/bin
+export GEM_PATH="$GEM_PATH:$gem_parent_dir"
+
+gem install foreman --install-dir "$gem_parent_dir" --bindir vendor/heroku/bin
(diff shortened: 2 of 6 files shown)


Procfile
```diff
@@ -1,2 +1,2 @@
-web: bundle exec foreman start -f Procfile.multi
+web: vendor/heroku/ruby/2.6.0/gems/foreman-0.85.0/bin/foreman start -f Procfile.multi
```
davidrunger: hardcoding the version number (`0.85.0`) here is brittle. something to improve later.

buildpack-run.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+gem_parent_dir="vendor/heroku/ruby/2.6.0"
+
+mkdir -p $gem_parent_dir
+mkdir -p vendor/heroku/bin
+export GEM_PATH="$GEM_PATH:$gem_parent_dir"
```
davidrunger: i'm not sure if this line is necessary?

Procfile
davidrunger: maybe use `vendor/heroku/bin` if possible; I think we currently aren't using it?

package.json
```diff
@@ -70,7 +70,7 @@
     "vue-drag-drop": "^1.1.4",
     "vue-form": "^4.10.1",
     "vue-loader": "^15.7.1",
-    "vue-router": "^3.1.2",
+    "vue-router": "^3.1.3",
```
davidrunger: this upgrade was sort of accidental ... but whatever

buildpack-run.sh
davidrunger: removed in https://github.com/davidrunger/david_runger/pull/1805

Procfile
davidrunger: fixed in https://github.com/davidrunger/david_runger/pull/1806 .

using `vendor/heroku/bin` doesn't work, so I removed that from `buildpack-run.sh` completely.

PR #1806, opened 2019-09-02
Don't refer to specific Ruby or foreman versions when running foreman

davidrunger:


Procfile
```diff
@@ -1,2 +1,2 @@
-web: vendor/heroku/ruby/2.6.0/gems/foreman-0.85.0/bin/foreman start -f Procfile.multi
+web: export GEM_PATH="$GEM_PATH:vendor/foreman" && vendor/foreman/bin/foreman start -f Procfile.multi
```
davidrunger: it's kind of ugly to make the command this long, but... 🤷‍♂ this seems to be necessary for it to work.

davidrunger: this version actually doesn't work ... I think we need to use the `foreman` CLI command from the foreman gem location, not from the bindir location

PR #1895, opened 2019-10-07
Tweak styling of text logs

davidrunger:
The text logs were being displayed in an ugly left-aligned way if the content of the log entries wasn't lengthy.


davidrunger:
This passed on Travis. Merging.

image

PR #1918, opened 2019-10-19
gzip responses (including HTML and JSON)

davidrunger:
It turns out that `rack-mini-profiler` changes the `Accept-Encoding` header to `"identity"` when active, so now we'll deactivate `rack-mini-profiler` by default (and instead require an `ENABLE_PROFILER` env variable in order to activate `rack-mini-profiler`) so that we'll have gzipping the rest of the time.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -56,7 +56,7 @@ group :development do
   gem 'listen', '>= 3.0.5', '< 3.3'
   # Performance profiling. Should be listed after `pg` (and `rails`?) gems to get database
   # performance analysis.
-  gem 'rack-mini-profiler'
+  gem 'rack-mini-profiler', require: false
   gem 'spring'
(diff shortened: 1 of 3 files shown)


davidrunger:
Testing against production (with 100 requests to `/api/log_entries` made via `curl` with a cookie header that had me logged in) I got `1.151 seconds` on average before deploy and `1.045 seconds` on average after deploy, i.e. ~`106 milliseconds` faster, and a response time decreased by ~9%. Not super amazing, but not too bad, either!

(Note that I have lots of log entries, so the response time (and absolute response time improvement) for me is probably greater than it would be for most users.)

davidrunger:
I just tried another 100 requests, and got `1.129 seconds`, which is only a 22ms improvement (and which also somewhat calls this whole methodology into question, in the first place).

PR #1920, opened 2019-10-19
Split env variable declarations in `.travis.yml` onto separate lines

davidrunger:
Not a substantive change; it just looks neater. :)


davidrunger:
Oh, woops, this triggers multiple builds, which isn't what I want:

image

Per https://docs.travis-ci.com/user/environment-variables/ : 
> One build will be triggered for each line in the `env` array.

and
> If you need to specify several environment variables for each build, put them all on the same line in the `env` array

Closing.

PR #1926, opened 2019-10-20
Use root_path as post-login redirect fallback (not groceries_path)

davidrunger:


davidrunger:
Woops, I submitted this PR with the wrong branch name.

PR #1928, opened 2019-10-20
Redirect to login_path when blocked by devise's `authenticate` method

davidrunger:
When a route is guarded by devise's `authenticate` method and a user is not logged in, devise redirects to `new_user_session_url`, if it exists. However, we have no such route (instead, we have `login_url`). This commit adds a `CustomFailureApp` (inheriting from `Devise::FailureApp`) that will redirect to that path (`login_path`) if a user tries to access an `authenticate` d route while not logged in.

### DIFF
--- a/app/lib/custom_failure_app.rb
+++ b/app/lib/custom_failure_app.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+class CustomFailureApp < Devise::FailureApp
+  def route(_scope)
+    :login_path
+  end
+end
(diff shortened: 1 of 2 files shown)


app/lib/custom_failure_app.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
```
davidrunger: This file path (`app/lib/custom_failure_app.rb`) is kind of weird (i.e. `app/lib/` rather than just `lib/`), but it's what is modeled in the devise docs, so whatever, I will follow their example.

PR #1936, opened 2019-10-20
Don't log about unpermitter parameters

davidrunger:
This is turned off by default in production, I think. Thus, this only affects `development` and `test`.

My primary motivation for turning this off is because is was logging stuff like `unpermitted_parameters=['request_uuid', 'request_uuid', 'request_uuid', 'request_uuid', 'request_uuid']`.

### DIFF
--- a/config/initializers/wrap_parameters.rb
+++ b/config/initializers/wrap_parameters.rb
@@ -8,6 +8,7 @@
 # Enable parameter wrapping for JSON. You can disable this by setting :format to an empty array.
 ActiveSupport.on_load(:action_controller) do
   wrap_parameters format: [:json]
+  ActionController::Parameters.action_on_unpermitted_parameters = false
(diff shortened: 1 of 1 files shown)


config/initializers/wrap_parameters.rb
```diff
@@ -8,6 +8,7 @@
 # Enable parameter wrapping for JSON. You can disable this by setting :format to an empty array.
 ActiveSupport.on_load(:action_controller) do
   wrap_parameters format: [:json]
+  ActionController::Parameters.action_on_unpermitted_parameters = false
```
davidrunger: I don't know if this is the best place to put this? It seems like a reasonably good place. I tried putting it at the "top level" in a (new) `config/initializers/action_controller.rb` file, but it didn't seem to have any effect there.

PR #1942, opened 2019-10-21
Add sidekiq-scheduler and add skeleton `Stats` worker

davidrunger:
The `Stats` worker can be used to send various metrics to StatsD. To start with, it just keeps track of how many users we have.

Another possible future use for `sidekiq-scheduler` might be to send myself an email summary of the requests that were made each day or something.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -151,6 +151,8 @@ GEM
       dotenv (= 2.7.5)
       railties (>= 3.2, < 6.1)
     erubi (1.9.0)
+    et-orbi (1.2.2)
+      tzinfo
     execjs (2.7.0)
     factory_bot (5.1.1)
       activesupport (>= 4.2.0)
@@ -168,6 +170,9 @@ GEM
       hashdiff
     flamegraph (0.9.5)
     formatador (0.2.5)
+    fugit (1.3.3)
+      et-orbi (~> 1.1, >= 1.1.8)
+      raabro (~> 1.1)
     globalid (0.4.2)
       activesupport (>= 4.2.0)
     guard (2.15.0)
@@ -298,6 +303,7 @@ GEM
       nio4r (~> 2.0)
     pundit (2.1.0)
       activesupport (>= 3.0.0)
+    raabro (1.1.6)
     rack (2.0.7)
     rack-mini-profiler (1.1.0)
       rack (>= 1.2.0)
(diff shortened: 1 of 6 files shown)


Gemfile.lock
```diff
@@ -404,6 +412,11 @@ GEM
       rack (>= 2.0.0)
       rack-protection (>= 2.0.0)
       redis (>= 4.1.0)
+    sidekiq-scheduler (3.0.0)
+      redis (>= 3, < 5)
+      rufus-scheduler (~> 3.2)
+      sidekiq (>= 3)
+      tilt (>= 1.4.0)
```
davidrunger: One thing I don't love is that the `sidekiq-scheduler` gem is a little dependency heavy. Memory is a valuable commodity, now that I am trying to run [3 processes](https://github.com/davidrunger/david_runger/blob/462bdb1e977935c55b49f88e9c76e6aad755e198/Procfile.multi#L1-L3) on a single 512 MB Heroku dyno.

If these dependencies use too much memory, then I might instead consider a much lighter approach, like this: https://medium.com/@jordinl/how-to-run-periodic-jobs-with-sidekiq-for-free-f031234c14a5

```rb
# scheduler_worker.rb
class SchedulerWorker
  include Sidekiq::Worker
  sidekiq_options queue: 'critical'

  SCHEDULE = {
    FirstWorker  => -> (time) { time.min == 5 },
    SecondWorker => -> (time) { time.min == 0 && time.hour == 9 },
    ThirdWorker  => -> (time) { time.min % 10 == 0 },
  }

  def perform
    execution_time = Time.zone.now
    execution_time -= execution_time.sec

    self.class.perform_at(execution_time + 60) unless scheduled?

    SCHEDULE.each do |(worker_class, schedule_lambda)|
      worker_class.perform_async if !scheduled?(worker_class) && schedule_lambda.call(execution_time)
    end
  end

  def scheduled?(worker_class = self.class)
    scheduled_workers[worker_class.name]
  end

  private

  def scheduled_workers
    @scheduled_workers ||= Sidekiq::ScheduledSet.new.entries.each_with_object({}) do |item, hash|
      hash[item['class']] = true
    end
  end
end
```

```rb
# sidekiq_initializer.rb
Sidekiq.configure_server do |config|
  config.on(:startup) do
    SchedulerWorker.perform_async unless SchedulerWorker.new.scheduled?
  end
end
```

PR #1944, opened 2019-10-21
Raise real Request::CreateRequestError error if saving request fails

davidrunger:
This will trigger Sidekiq retries, which we now have! :)

### DIFF
--- a/app/workers/save_request.rb
+++ b/app/workers/save_request.rb
@@ -23,9 +23,20 @@ class SaveRequest
       return
     end
 
-    saved_successfully = request.save
-    if !saved_successfully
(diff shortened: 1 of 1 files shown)


app/workers/save_request.rb
```diff
@@ -102,21 +113,4 @@ def warn_about_missing_final_stashed_json
       request_uuid: @request_uuid,
     )
   end
-
-  def warn_about_failure_to_save_request
```
davidrunger: (Sending the warning to Rollbar manually is no longer necessary; Rollbar should send the raised error automatically.)

PR #1955, opened 2019-10-22
Add requests.request_id column (and populate it for new requests)

davidrunger:
This will facilitate connecting logs lines to `Request` records, which might be useful!

### DIFF
--- a/app/models/request.rb
+++ b/app/models/request.rb
@@ -5,7 +5,7 @@
 # Table name: requests
 #
 #  db           :integer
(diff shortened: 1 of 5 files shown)


app/models/request.rb
```diff
@@ -5,7 +5,7 @@
 # Table name: requests
 #
 #  db           :integer
-#  format       :string           not null
```
davidrunger: (unrelated `annotate` fix/change from running `annotate --models`)

PR #1972, opened 2019-10-26
Use a Redis connection pool (rather than Redis instance directly)

davidrunger:
The two naive approaches for connecting to a database (Redis, in this case) are:
1. just use a single global Redis connection for the whole Rails app, e.g. `$redis = Redis.new` in an initialzier and then `$redis.get('some_key')` in application code; this was our approach prior to this commit
2. initialize a new Redis connection each time that we need to interact with Redis (e.g. `Redis.new.get('some_key')` in application code)

---

The disadvantage of #1 is that Thread B might be blocked waiting for the connection, if the single, global connection is already in use by Thread A:

> at best, your threads will block while they wait for the connection to be free, and at worst, one thread will receive the results of another thread's query!

-- https://mailchi.mp/railsspeed/should-you-be-using-puma-with-multiple-threads?e=cd6e8b0ab8

---

The disadvantage of #2 is that it wastes time establishing the connection:

> Opening and maintaining a database connection for each user, especially requests made to a dynamic database-driven website application, is costly and wastes resources.

-- https://en.wikipedia.org/w/index.php?title=Connection_pool&oldid=913204739

---

The approach implemented in this PR -- connection pooling -- achieves the best of both worlds by creating multiple connections (on demand) but then reusing them after creation.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,6 +12,7 @@ end
 gem 'active_model_serializers', github: 'rails-api/active_model_serializers', branch: '0-10-stable'
 gem 'administrate', github: 'thoughtbot/administrate' # source from master for Rails 6 compatibility
 gem 'browser'
+gem 'connection_pool'
 gem 'devise'
 gem 'dotenv-rails', require: 'dotenv/rails-now'
 gem 'hamlit'
--- a/config/initializers/redis.rb
+++ b/config/initializers/redis.rb
@@ -1,3 +1,4 @@
 # frozen_string_literal: true
 
-$redis = Redis.new
+rails_max_threads = Integer(ENV.fetch('RAILS_MAX_THREADS') { 5 })
+$redis_pool = ConnectionPool.new(size: rails_max_threads, timeout: 1) { Redis.new }
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -465,6 +465,7 @@ DEPENDENCIES
   binding_of_caller
   browser
   capybara
+  connection_pool
   devise
   dotenv-rails
   factory_bot_rails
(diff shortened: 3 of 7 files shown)


Gemfile
```diff
@@ -12,6 +12,7 @@ end
 gem 'active_model_serializers', github: 'rails-api/active_model_serializers', branch: '0-10-stable'
 gem 'administrate', github: 'thoughtbot/administrate' # source from master for Rails 6 compatibility
 gem 'browser'
+gem 'connection_pool'
```
davidrunger: (This is already a dependency via Sidekiq, but since we are now using it explicitly ourselves, I am adding it to the `Gemfile` now.)

config/initializers/redis.rb
```diff
@@ -1,3 +1,4 @@
 # frozen_string_literal: true
 
-$redis = Redis.new
+rails_max_threads = Integer(ENV.fetch('RAILS_MAX_THREADS') { 5 })
+$redis_pool = ConnectionPool.new(size: rails_max_threads, timeout: 1) { Redis.new }
```
davidrunger: This timeout (1 second) seems fairly aggressive, but I am sort of curious to see if we will ever hit it when set this aggressively. Although it seems a little aggressive, on the one hand, on the other hand it also seems like we don't want to be waiting even 1 second just to get a Redis connection.

davidrunger:
This PR (maybe combined with `ab -n 10 -c 10 'https://www.davidrunger.com/'`, which I ran 3 times shortly after deploy) caused a bunch of `Redis::CommandError: ERR max number of clients reached` errors.

davidrunger:
> I need to decide how to split those connections between Sidekiq itself and the Rails server.
>
> I guess that this work also should be considered along with my other planned work to decrease the number of Sidekiq threads and/or Rails server threads/processes that I am running, to decrease memory usage.

I am indeed tackling all of these considerations together in https://github.com/davidrunger/david_runger/pull/1973/ (_Reduce the number of threads, processes, & Redis connections used_).

PR #1974, opened 2019-10-26
Add Sidekiq stats tracking, and `essential` Sidekiq queue

davidrunger:
The `essential` queue is for monitoring jobs that we don't want to get blocked by other processes. (Ideally we'd run this queue on an entirely separate Sidekiq process, but we can't spare another 6+ Redis connections.)

All jobs executed in this queue should be expected to execute quickly. I might look into whether there is some way that I can kill any jobs that take longer than 5 seconds or something.

### DIFF
--- a/config/sidekiq.yml
+++ b/config/sidekiq.yml
@@ -2,8 +2,15 @@
 production:
   :concurrency: 2
 
+:queues:
+  - ['essential', 12] # ideally we'd run this queue on a separate process, but that'd cost money :(
+  - ['default', 3]
+  - ['mailers', 1]
+
 :schedule:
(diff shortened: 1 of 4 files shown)


config/sidekiq.yml
```diff
@@ -2,8 +2,15 @@
 production:
   :concurrency: 2
 
+:queues:
+  - ['essential', 12] # ideally we'd run this queue on a separate process, but that'd cost money :(
```
davidrunger: I forgot to actually run this queue. 🙃 Fixed in https://github.com/davidrunger/david_runger/pull/1975 .

PR #1979, opened 2019-10-27
Add Sidekiq middleware in development with more detailed logging

davidrunger:
It looks like this:
image

It logs `[queue] [job class] [jid (job id)] [args]`, and also clear indicators for when the job starts and finishes. (The ActiveRecord logging seen in this screenshot is added elsewhere, in a git-ignored file.)

### DIFF
--- a/config/initializers/sidekiq.rb
+++ b/config/initializers/sidekiq.rb
@@ -32,6 +32,15 @@ Sidekiq.configure_server do |config|
     config.server_middleware do |chain|
       require_relative '../../lib/middleware/set_config_sidekiq_middleware'
       chain.add(Middleware::SetConfigSidekiqMiddleware)
+
+      require_relative '../../lib/middleware/sidekiq_dev_logging'
+      chain.add(Middleware::SidekiqDevLogging)
+      # silence the default Sidekiq logger
+      class ::Sidekiq::JobLogger
+        def call(*_args)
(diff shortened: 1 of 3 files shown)


lib/middleware/set_config_server_middleware.rb
```diff
@@ -1,5 +1,9 @@
 # frozen_string_literal: true
 
+# I'm not sure why it's necessary to explicitly define this module; I thought zeitwerk is supposed
+# to do it for us. Maybe it's because it gets loaded so early.
+module Middleware; end
```
davidrunger: Unrelated change. When I was trying to boot `bin/rails server` locally, it was complaining about a missing `Middleware` constant? Whatever. I'm adding it here.

config/initializers/sidekiq.rb
```diff
@@ -32,6 +32,15 @@
     config.server_middleware do |chain|
       require_relative '../../lib/middleware/set_config_sidekiq_middleware'
       chain.add(Middleware::SetConfigSidekiqMiddleware)
+
+      require_relative '../../lib/middleware/sidekiq_dev_logging'
+      chain.add(Middleware::SidekiqDevLogging)
```
davidrunger: (note that this code change is within an `if Rails.env.development?` block)

PR #1980, opened 2019-10-27
Track Redis db0 & db1 stats even if they don't have any keys

davidrunger:
When a Redis database doesn't have any keys, it isn't included in the stats returned by the `INFO` command. If this happens, we want to track that there are 0 keys, rather than omitting the stats entirely.

### DIFF
--- a/app/workers/redis_stats.rb
+++ b/app/workers/redis_stats.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 class RedisStats
+  extend Memoist
   include Sidekiq::Worker
 
   sidekiq_options(queue: :essential)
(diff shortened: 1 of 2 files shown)


app/workers/redis_stats.rb
```diff
@@ -57,29 +58,32 @@ class RedisStats
   ].freeze
 
   def perform
-    @info_hash = $redis_pool.with(&:info)
```
davidrunger: I switched from an instance variable to a method in order to make stubbing easier in tests.

app/workers/redis_stats.rb
```diff
@@ -57,29 +58,32 @@ class RedisStats
...
-
     track_integers
     track_floats
     track_databases
     track_connection_pool
   end
 
-  private
+  private unless Rails.env.test?
```
davidrunger: This is kind of a neat(/somewhat surprising) feature of Ruby that the `private` keyword can be conditional. I like to do this sometimes to make it easier to unit test private methods.

PR #1983, opened 2019-10-27
Include IP address in web logs

davidrunger:
This makes it easier to such across multiple requests from a single IP if there isn't a user logged in (and especially if the machine/program making the request doesn't store cookies).

### DIFF
--- a/lib/david_runger/log_builder.rb
+++ b/lib/david_runger/log_builder.rb
@@ -13,7 +13,8 @@ class DavidRunger::LogBuilder
       exception: exception_log,
       params: params_log,
       status: status_code_log,
-      user_id: user_id_log,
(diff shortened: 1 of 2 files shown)


lib/david_runger/log_builder.rb
```diff
@@ -46,8 +47,4 @@ def status_code_log
     # Devise annoyingly sets payload[:status] to 401 sometimes even when an exception has occurred
     exception.present? ? 500 : payload[:status]
   end
-
-  def user_id_log
-    payload[:user_id] # comes from ApplicationController#append_info_to_payload
-  end
```
davidrunger: more-or-less unrelated refactor; it seemed excessive to have a separate method for doing this

PR #1986, opened 2019-10-28
Create generic PapertrailSearchField & use for request_id

davidrunger:
(... on the Request Admin show page)


app/views/fields/papertrail_search_field/_index.html.erb
```diff
@@ -0,0 +1 @@
+<%= field.to_s %>
```
davidrunger: (This view isn't yet used (since we don't show `request_id` in the request index page, and `request_id` is the only field currently using this `PapertrailSearchField` thus far), but we should probably add it, anyway.)

PR #1987, opened 2019-10-28
Only stash request data in Redis if we will log a Request

davidrunger:
The criterion that we are using to determine whether we will log a Request is simply whether or not the controller in question inherits from `ApplicationController`. This excludes requests to PgHero pages and to Admin pages, for example.

### DIFF
--- a/app/controllers/concerns/request_recordable.rb
+++ b/app/controllers/concerns/request_recordable.rb
@@ -11,7 +11,7 @@ module RequestRecordable
 
   included do
     prepend_before_action :set_request_time
-    before_action :store_initial_request_data_in_redis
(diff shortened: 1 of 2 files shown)


app/controllers/concerns/request_recordable.rb
```diff
@@ -11,7 +11,7 @@ class StashRequestError < StandardError; end
 
   included do
     prepend_before_action :set_request_time
-    before_action :store_initial_request_data_in_redis
+    before_action :store_initial_request_data_in_redis, if: :inherits_from_application_controller?
```
davidrunger: I realized that this check isn't actually necessary. I've submitted https://github.com/davidrunger/david_runger/pull/1988 to remove it.

PR #2003, opened 2019-11-02
Avoid yarn installing twice during deploys

davidrunger:
Also, remove `clean_yarn_cache` and `rmrf_node_module` rake task, since it sounds like those steps might not be necessary any more (from the discussion at https://github.com/yarnpkg/yarn/issues/ 1832 )?

### DIFF
--- a/lib/tasks/assets.rake
+++ b/lib/tasks/assets.rake
@@ -10,19 +10,9 @@ namespace :assets do
     end
   end
 
-  desc 'clean yarn cache'
-  task :clean_yarn_cache do
-    run_logged_system_command('yarn cache clean')
-  end
-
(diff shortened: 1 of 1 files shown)


lib/tasks/assets.rake
```diff
@@ -174,12 +167,16 @@ module SourceMapHelper
 end
 
 if Rails.env.production?
-  Rake::Task['assets:precompile'].enhance(%w[
-    assets:clean_yarn_cache
-    assets:rmrf_node_module
```
davidrunger: these rake tasks were originally added in https://github.com/davidrunger/david_runger/commit/9bd168a888add2b66dd379a3d51c939a7737cafb

PR #2004, opened 2019-11-02
Avoid yarn installing twice during deploys (second try)

davidrunger:
This builds on https://github.com/davidrunger/david_runger/pull/2003 . Hopefully this attempt will actually work. I think that the problem with that attempt is that the `assets:precompile` rake task hadn't yet been enhanced with `yarn:install` at the time that we tried to remove `yarn:install` from its prerequisites. I got the hint to do it in `Rakefile` from https://github.com/heroku/heroku-buildpack-ruby/issues/ 654#issuecomment-387226196 .

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -6,3 +6,11 @@
 require_relative 'config/application'
 
 Rails.application.load_tasks
+
+if Rails.env.production? && DeployAssetsHelper.on_heroku?
+  # https://github.com/heroku/heroku-buildpack-ruby/pull/892#issuecomment-548897899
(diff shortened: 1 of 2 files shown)


davidrunger:
Nice, this worked!

image

It seems to have shaved ~40 seconds off of the deploy time, which is about what I'd expect, since that is how long each of the two (now one) `yarn install`s were taking.

The build/deploy time (at least when dependencies haven't been changed) is now split ~ evenly between 4 minutes of running the tests and 4 minutes of deploying to Heroku.

PR #2093, opened 2019-12-07
Update robots.txt to explicitly allow bots to access the whole site

davidrunger:
(An alternative to `Disallow:` is `Allow: /`, but `Disallow:` seems more widely supported.)


public/robots.txt
```diff
@@ -1 +1,2 @@
-# See http://www.robotstxt.org/robotstxt.html for documentation on how to use the robots.txt file
+User-agent: *
+Disallow:
```
davidrunger: (The empty `Disallow` value means "allow all".)

PR #2094, opened 2019-12-08
Return 404 rather than 500 when route doesn't match

davidrunger:
Also, fix logging of request method and path for 404s. (I'm not sure what exactly changed, or when...)


davidrunger:
One of the bugs fixed by this PR was probably introduced some time around 2019-10-26.

davidrunger:
The bug was probably introduced in https://github.com/davidrunger/david_runger/pull/1981 .

PR #2139, opened 2019-12-21
Don't install development or production gems on Travis

davidrunger:
Maybe this will make Travis runs sometimes execute a little bit faster? Anyway, isn't this the point of having a `test` group in the `Gemfile`?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,9 @@ before_install:
   - CHROMEDRIVER_VERSION=$(wget -qO- https://chromedriver.storage.googleapis.com/LATEST_RELEASE)
   - curl -o tmp/chromedriver.zip https://chromedriver.storage.googleapis.com/$CHROMEDRIVER_VERSION/chromedriver_linux64.zip
   - unzip -d "$HOME/bin/" tmp/chromedriver.zip
+install:
+  - bundle config set without 'development:production'
+  - bundle install --jobs=3 --retry=3
 script:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -22,6 +22,7 @@ before_install:
   - CHROMEDRIVER_VERSION=$(wget -qO- https://chromedriver.storage.googleapis.com/LATEST_RELEASE)
   - curl -o tmp/chromedriver.zip https://chromedriver.storage.googleapis.com/$CHROMEDRIVER_VERSION/chromedriver_linux64.zip
   - unzip -d "$HOME/bin/" tmp/chromedriver.zip
+install: bundle install --without development:production --jobs=3 --retry=3
```
davidrunger: relevant docs: https://docs.travis-ci.com/user/languages/ruby/

---

> Travis CI uses Bundler to install your Ruby project’s dependencies if there is a Gemfile in the project’s root directory, or if there is a Gemfile specified in the build matrix:

```
bundle install --jobs=3 --retry=3
```

If a Gemfile.lock exists in your project’s root directory, we add the `--deployment` flag.

If you want to use a different means of handling your Ruby project’s dependencies, you can override the install command.

```yaml
# .travis.yml
install: gem install rails
```

davidrunger: the `--without development:production` syntax is deprecated, so i added a `bundle config set without 'development:production'` command instead.

PR #2140, opened 2019-12-21
Move some gems (dotenv-rails, rubocop, etc) from default group

davidrunger:
Since we don't need these gems in production, we shouldn't load/install them in production! This commit moves the gems to the `:development, :test` group.


davidrunger:
This change reduced the Heroku slug size from `149.4M` to `148M`. That's not nothing! :)

PR #2141, opened 2019-12-21
Move all tasks previously managed by Heroku Scheduler to Sidekiq

davidrunger:
It's better to have things defined in publicly accessible, git-tracked code than hidding in config settings without any history.

### DIFF
--- a/lib/tasks/db.rake
+++ b/lib/tasks/db.rake
@@ -8,7 +8,7 @@ module DavidRunger::TruncateTables
   SQL
 
   def self.max_allowed_rows
-    Integer(ENV['MAX_TABLE_ROWS']) || 2_000
+    Integer(ENV.fetch('MAX_TABLE_ROWS') { 2_000 })
   end
 
   def self.print_row_counts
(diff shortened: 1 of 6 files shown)


davidrunger:
We could add less code by having a generic `InvokeRakeTask` worker or something, and passing the rake task name as an argument, but I don't know if that's better or not. It wouldn't be so clear in the logs which work is being done, then. I think that, for now, it makes sense to have a separate worker for each different rake task.

lib/tasks/db.rake
```diff
@@ -8,7 +8,7 @@ module DavidRunger::TruncateTables
   SQL
 
   def self.max_allowed_rows
-    Integer(ENV['MAX_TABLE_ROWS']) || 2_000
+    Integer(ENV.fetch('MAX_TABLE_ROWS') { 2_000 })
```
davidrunger: mostly unrelated change; this was erroring on my local machine during testing because I don't have a `MAX_TABLE_ROWS` env variable set.

PR #2148, opened 2019-12-23
Immediately write location & ISP data for each request's IP address

davidrunger:


spec/spec_helper.rb
```diff
@@ -15,6 +15,7 @@
 require 'rspec/rails'
 require 'capybara/rails'
 require 'capybara/rspec'
+require 'active_support/cache/dalli_store'
```
davidrunger: I tried to put this `require` alphabetically (first), but an error was raised when doing so, which moving it to the end here avoids.

.travis.yml
```diff
@@ -7,6 +7,7 @@ cache:
 rvm:
   - "2.6.5"
 services:
+  - memcached
```
davidrunger: this is needed since we are herein adding our first test that relies on cacheing.

PR #2170, opened 2020-01-03
Check only bundler major and minor version numbers (not patch)

davidrunger:
This will avoid failures if/when Travis randomly bumps the bundler patch version.


davidrunger:
This PR might partially be necessitated by this change, in which we stopped installing a specific version of bundler https://github.com/davidrunger/david_runger/commit/e0c56bafc9008b0a12497e034194202e96ae4600#diff-354f30a63fb0907d4ad57269548329e3 ?

PR #2198, opened 2020-01-07
Call `Oj.optimize_rails` in a (new) initializer, to actually use Oj

davidrunger:
I realized from this https://github.com/evilmartians/terraforming-rails/tree/master/tools/gem_tracker tool that I wasn't actually using the `oj` gem.

Ironically, actually, that tool didn't really prove that I wasn't using `oj`, because -- since `oj` is a C extension -- that tool makes it look like `oj` isn't being used even when it _is_ being used.

However, the fact that the tool said that `oj` isn't being used got me to investigate, and it turns out that, indeed, we weren't using `oj`. But now we are!, and we will enjoy faster JSON serialization because of it.

### DIFF
--- a/config/initializers/oj.rb
+++ b/config/initializers/oj.rb
@@ -1,0 +1,3 @@
+# frozen_string_literal: true
+
+Oj.optimize_rails


davidrunger:
Some crude measurements suggest that this might have improved the performance of `Api::LogEntriesController#index` (which is probably our most JSON-serialization-heavy endpoint) by about 10% (which is also about what eyeballing the NewRelic graph below suggests to me).

image

-- https://rpm.newrelic.com/accounts/1627906/applications/59388619/transactions#id=5b22436f6e74726f6c6c65722f6170692f6c6f675f656e74726965732f696e646578222c22225d

PR #2211, opened 2020-01-11
Center the success/failure messages about sending a groceries text

davidrunger:


app/javascript/groceries/components/store.vue
```diff
@@ -263,4 +265,11 @@ export default {
     background: #219b21;
   }
 }
+
+// double the `.toastify-center` class to ensure it has higher precedence than the library's CSS
+.toastify-center.toastify-center {
+  left: inherit;
+  right: 50%;
+  transform: translateX(50%);
```
davidrunger: this all is necessary because the library's CSS makes a centered toast message be ~50% of the width of the viewport, rather than being only as wide as the text content, which is what I want

PR #2213, opened 2020-01-12
Style the "Sign in with Google" button

davidrunger:
One might have hoped that this could somehow be simpler. Maybe it could be...

I basically tried to steal the styling from here: https://developers.google.com/identity/sign-in/web/build-button

---

image

### DIFF
--- a/app/views/sessions/new.html.erb
+++ b/app/views/sessions/new.html.erb
@@ -1,0 +1,225 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <title>Login - David Runger</title>
+    <style type="text/css">
+      /* CSS stolen from Google (with formatting tweaks and others)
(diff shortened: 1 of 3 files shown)


app/views/sessions/new.html.erb
```diff
@@ -0,0 +1,225 @@
...
+    <script>
+      function submitLoginForm(event) {
+        event.preventDefault();
+
+        const form = document.querySelector('form[action="/users/auth/google_oauth2"]');
+        form.submit();
+      }
+
+      const googleLoginButton = document.getElementById('my-signin2');
+      googleLoginButton.addEventListener('click', submitLoginForm, false);
```
davidrunger: we submit the form via JavaScript rather than just via clicking the submit button in order to make the login work no matter what part of the whole login-with-Google button/image the user clicks on. without this, the user would have to click directly on the "Sign in with Google" text.

davidrunger:
> One might have hoped that this could somehow be simpler. Maybe it could be...

I think that a simpler way to do this would be to use an image rather than rendering the button with HTML/CSS.

PR #2243, opened 2020-01-26
Allow users to upload log data via a CSV

davidrunger:


config/routes.rb
```diff
@@ -20,7 +20,10 @@
   get 'groceries', to: 'groceries#index'
 
   get 'logs', to: 'logs#index'
-  get 'logs/:slug', to: 'logs#index' # routing to specific log will be done by Vue Router
+  namespace :logs do
+    resources :uploads, only: %i[index create]
+  end
+  get 'logs/:slug', to: 'logs#index', as: :log # routing to specific log will be done by Vue Router
```
davidrunger: (The order here matters; this needs to come below `namespace :logs [...] resources :uploads`; otherwise, `uploads` will just be interpreted as a `:slug`.)

lib/tasks/db.rake
```diff
@@ -8,7 +8,7 @@ module DavidRunger::TruncateTables
   SQL
 
   def self.max_allowed_rows
-    Integer(ENV.fetch('MAX_TABLE_ROWS') { 2_000 })
+    Integer(ENV.fetch('MAX_TABLE_ROWS') { 1_500 })
```
davidrunger: I've updated this ENV variable in Heroku already. I'm making this change because I'll be adding a lot more rows to the database, and I only have I think 10,000 rows on Heroku's free Postgres database plan, so I need to free up some rows.

PR #2249, opened 2020-01-26
Add rubycritic gem

davidrunger:
We won't run this on CI just yet. For now, the purpose of adding this gem is to allow executing `rubycritic` locally:
```
bundle exec rubycritic `find app -type d -not -name dashboards -d 1` lib
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -53,6 +53,7 @@ group :development, :test do
   gem 'rubocop-performance', require: false
   gem 'rubocop-rails', require: false
   gem 'rubocop-rspec', require: false
+  gem 'rubycritic', require: false, github: 'whitesmith/rubycritic'
 end
 
 group :development do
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -53,6 +53,7 @@ group :development, :test do
   gem 'rubocop-performance', require: false
   gem 'rubocop-rails', require: false
   gem 'rubocop-rspec', require: false
+  gem 'rubycritic', require: false, github: 'whitesmith/rubycritic'
```
davidrunger: We need to install from GitHub to get this recent change: https://github.com/whitesmith/rubycritic/pull/ 342

Gemfile.lock
```diff
@@ -421,7 +468,7 @@ GEM
       rufus-scheduler (~> 3.2)
       sidekiq (>= 3)
       tilt (>= 1.4.0)
-    simplecov (0.16.1)
+    simplecov (0.17.1)
```
davidrunger: Installing the gem successfully required `bundle update simplecov`.

PR #2269, opened 2020-02-02
Simplify Api::LogEntriesController#index action

davidrunger:
Previously, we were returning differently formatted JSON data, depending on whether the request was for the log entries of a single log or the log entries of all of the current user's logs.

This change simplifies the controller action by always returning the log entries data in the same format, regardless of whether we are returning the logs entries of one log or many logs.

This basically just moves the complexity from the backend to the frontend... but that feels slightly better. This complexity sort of feels like something that the frontend should manage.

### DIFF
--- a/app/controllers/api/log_entries_controller.rb
+++ b/app/controllers/api/log_entries_controller.rb
@@ -37,29 +37,22 @@ class Api::LogEntriesController < ApplicationController
   def index
     log_id = params['log_id']
 
-    if log_id.present?
-      render json: ActiveModel::Serializer::CollectionSerializer.new(
-        current_user.logs.find(log_id).log_entries_ordered,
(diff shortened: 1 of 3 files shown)


davidrunger:
I think that this change probably entails a performance penalty, for at least two reasons:
1. we are now doing some extra work on the client side that we weren't doing before
2. the response payload when requesting all logs will be larger

I don't really care, though. I'm guessing that the performance penalty is negligible. Additionally, this scenario happens "in the background", anyway, basically proactively cacheing all of the user's log data locally before they need it, so performance isn't a major concern (unless the slowdown is significant enough that the cacheing now doesn't finish before the user needs the data, i.e. before the user views their first log or a different log than the initial log).

PR #2270, opened 2020-02-02
Enable Sidekiq (testing SaveRequest, etc.) in home page feature test

davidrunger:
This gives us some integration test coverage for the `SaveRequest` and `FetchIpInfoForRequest` Sidekiq workers.

### DIFF
--- a/app/workers/fetch_ip_info_for_request.rb
+++ b/app/workers/fetch_ip_info_for_request.rb
@@ -29,7 +29,7 @@ class FetchIpInfoForRequest
 
   memoize \
   def ip_info_from_api(ip)
-    logger.info("Querying ip-api.com for info about IP address '#{ip}'")
+    Rails.logger.info("Querying ip-api.com for info about IP address '#{ip}'")
     raw_ip_info_from_api = HTTParty.get("http://ip-api.com/json/#{ip}").parsed_response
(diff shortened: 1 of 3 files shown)


app/workers/fetch_ip_info_for_request.rb
```diff
@@ -29,7 +29,7 @@ def cache_key(ip)
 
   memoize \
   def ip_info_from_api(ip)
-    logger.info("Querying ip-api.com for info about IP address '#{ip}'")
+    Rails.logger.info("Querying ip-api.com for info about IP address '#{ip}'")
```
davidrunger: I'm changing this from `logger` to `Rails.logger` because the former's printed output was being logged to my terminal when running tests.

spec/spec_helper.rb
```diff
@@ -24,6 +24,7 @@
 require 'capybara/rails'
 require 'capybara/rspec'
 require 'active_support/cache/dalli_store'
+require 'sidekiq/testing'
```
davidrunger: This is necessary to be able to call `Sidekiq::Testing.inline!` in specs; otherwise, we get an error about an undefined constant `Sidekiq::Testing.`.

davidrunger:
Enabling Sidekiq in this one feature spec is giving us a `+3.05%` gain in overall test coverage!

PR #2271, opened 2020-02-02
Add unit test for TouchActivityAt worker

davidrunger:


app/workers/touch_activity_at.rb
```diff
@@ -7,7 +7,7 @@ def perform(user_id, request_time_as_float)
     user = User.find_by(id: user_id)
 
     if user.blank?
-      logger.info("User #{user_id} was not found to update last_activity_at")
+      Rails.logger.info("User #{user_id} was not found to update last_activity_at")
```
davidrunger: I'm switching from `logger` to `Rails.logger` here so that it's easier to reference the logger in specs.

PR #2317, opened 2020-02-22
Add Skylight

davidrunger:
(Note: I also set the `SKYLIGHT_AUTHENTICATION` ENV variable via the Heroku web UI.)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -34,6 +34,7 @@ gem 'redis', '~>4.1'
 gem 'rollbar'
 gem 'sidekiq'
 gem 'sidekiq-scheduler', require: false # required manually in config/initializers/sidekiq.rb
+gem 'skylight'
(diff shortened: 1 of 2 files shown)


davidrunger:
According to NewRelic, it appears that Skylight is slowing down our `HomeController#index` response times somewhat:

image

# Before (detail)

image

# After (detail)

image

davidrunger:
Because of the aforementioned performance issue I've disabled Skylight for now with `heroku config:set SKYLIGHT_ENABLED=false`.

I'll leave Skylight in the codebsae, though, so that I can easily turn it back on with `heroku config:set SKYLIGHT_ENABLED=true` if I need to debug a performance issue

davidrunger:
It looks like disabling Skylight has indeed restored performance to the previous level:
image

PR #2319, opened 2020-02-22
Only show top-left "David Runger" 'logo' after scrolling down

davidrunger:
It otherwise looks redundant to have two "David Runger" texts.

# Before

image

# After

image


spec/javascript/home/home.spec.js
```diff
@@ -1,18 +0,0 @@
-import 'spec_helper';
-import { createLocalVue, mount } from '@vue/test-utils';
-import Home from 'home/home.vue';
-
-describe('Home', function () { // eslint-disable-line func-names, prefer-arrow-callback
```
davidrunger: I am deleting this test because it was [failing](https://travis-ci.org/davidrunger/david_runger/builds/653894320) and it's basically redundant with [this feature test](https://github.com/davidrunger/david_runger/blob/70cf4aa116ca3dfab453dd20033f1996078236a1/spec/features/home_spec.rb#L59-L62), anyway.

Overall, I'm not too happy with the Vue testing ecosystem yet. I appreciate the work/effort that has been put in to it, but it just hasn't worked out all too well for me so far.

PR #2453, opened 2020-04-11
Add quick-submit buttons for recently submitted counter log values

davidrunger:
Clicking these buttons is easier than typing in the number.

We'll show quick-buttons for (at most) each of the most 5 recently submitted values for the log being viewed.

image

### DIFF
--- a/app/javascript/logs/components/new_log_entry_form.vue
+++ b/app/javascript/logs/components/new_log_entry_form.vue
@@ -1,6 +1,16 @@
 <template lang='pug'>
 div
-  vue-form.px1(@submit.prevent='postNewLogEntry' :state='formstate' :class='log.data_type')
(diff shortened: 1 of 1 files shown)


app/javascript/logs/components/new_log_entry_form.vue
```diff
@@ -11,7 +21,7 @@ div
         :type='inputType'
       )
     el-input.new-log-input.mb1(
-      v-if='isDuration || isNumber'
+      v-if='isCounter || isDuration || isNumber'
```
davidrunger, line 24: here we allow adding notes for counter logs. currently, the notes won't be shown in the UI anywhere. however, we might later add functionality to show the notes in the UI, so we'll start collecting the data if the user wants to submit it.

PR #2519, opened 2020-05-02
Remove graphite deploy-notification release command

davidrunger:
The graphite deploy-notification release command is causing deploys to fail, since I deleted the Digital Ocean droplet last night.


davidrunger:
This undoes much of the changes of https://github.com/davidrunger/david_runger/pull/414 .

PR #2567, opened 2020-05-16
Update all gems

davidrunger:
`bundle update`


davidrunger:
### Motivation

In https://github.com/davidrunger/david_runger/pull/2566/ , we fixed an issue with `guard`/`pry` incompatibility by bumping `guard`. Are there other beneficial changes that we are missing out on from other outdated gems? Let's assure ourselves that the answer is as close to "no" as possible, via this PR, which updates as many gems as possible (via `bundle update`).

PR #2568, opened 2020-05-16
Update all JavaScript packages

davidrunger:
`yarn upgrade`

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -77,26 +77,18 @@
     semver "^5.5.0"
 
 "@babel/helper-create-class-features-plugin@^7.8.3":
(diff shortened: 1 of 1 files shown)


davidrunger:
### Motivation

In https://github.com/davidrunger/david_runger/pull/2566/ , we fixed an issue with `guard`/`pry` incompatibility by bumping `guard`. Are there other beneficial changes that we are missing out on from outdated JavaScript packages? Maybe. This PR aims to reduce the likelihood of that.

Since we're talking about JavaScript here, though, it seems just as likely that this might cause problems... but let's give it a shot.

PR #2570, opened 2020-05-16
Test Api::LogsController#update

davidrunger:


spec/controllers/api/logs_controller_spec.rb
```diff
@@ -34,6 +34,7 @@
           'description' => nil,
           'id' => Integer,
           'name' => 'Resting Heart Rate',
+          'publicly_viewable' => false,
```
davidrunger, line 37: (unrelated change, just noting the addition of `publicly_viewable` in `#create`)

PR #2571, opened 2020-05-16
Remove pghero

davidrunger:
It was good while it lasted.

fixes https://rollbar.com/davidjrunger/davidrunger/items/220/

1. prints Ruby 2.7 warnings
2. not compatible with `rails` master (see Rollbar above)
3. not worth it; DB performance isn't a big problem for the app
4. tracking tables use/waste rows in database

### DIFF
--- a/db/migrate/20200516091729_disable_pg_stat_statements_extension.rb
+++ b/db/migrate/20200516091729_disable_pg_stat_statements_extension.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+class DisablePgStatStatementsExtension < ActiveRecord::Migration[6.1]
+  def up
+    ApplicationRecord.connection.execute('DROP EXTENSION IF EXISTS pg_stat_statements CASCADE;')
+  end
+
+  def down
+    ApplicationRecord.connection.execute('CREATE EXTENSION IF NOT EXISTS pg_stat_statements;')
+  end
+end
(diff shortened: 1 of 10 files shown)


db/migrate/20200516091729_disable_pg_stat_statements_extension.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class DisablePgStatStatementsExtension < ActiveRecord::Migration[6.1]
+  def up
+    ApplicationRecord.connection.execute('DROP EXTENSION IF EXISTS pg_stat_statements CASCADE;')
```
davidrunger, line 5: (The `CASCADE` also drops the `pghero_query_stats` and `pghero_space_stats` tables.)

db/migrate/20200516091729_disable_pg_stat_statements_extension.rb
```diff
@@ -0,0 +1,11 @@
...
+  end
+
+  def down
+    ApplicationRecord.connection.execute('CREATE EXTENSION IF NOT EXISTS pg_stat_statements;')
```
davidrunger, line 9: (to properly reverse the `#up`, I should probably create the `pghero_query_stats` and `pghero_space_stats` tables, but I don't think it's worth doing / pushing a fix for.)

PR #2583, opened 2020-05-18
Improve database/models per `database_consistency` recommendations

davidrunger:
Also, add `bundle exec database_consistency` as a check in Travis, to ensure that future changes comply with `database_consistency`'s recommendations (which seem to be good/accurate 👍).

These were the suggestions/comments:
```
fail User email column is required in the database but do not have presence validator
fail Item id column has int/serial type but recommended to have bigint/bigserial/uuid
fail Log slug column is required in the database but do not have presence validator
fail Store id column has int/serial type but recommended to have bigint/bigserial/uuid
fail User sms_records associated model should have proper index in the database
fail User stores associated model should have proper index in the database
fail Log number_log_entries associated model should have proper index in the database
fail Log number_log_entries_ordered associated model should have proper index in the database
fail Log text_log_entries associated model should have proper index in the database
fail Log text_log_entries_ordered associated model should have proper index in the database
fail Request status column should be required in the database
fail Request request_id column should be required in the database
```


.travis.yml
```diff
@@ -31,11 +31,12 @@ script:
   - chromedriver --version # print chromedriver version, but don't check it, because it changes
   - yarn install
   - bin/rails spec:setup_js > /dev/null 2>&1 &
-  - bundle exec rubocop $(git ls-tree -r HEAD --name-only) --force-exclusion --format clang
-  - bundle exec brakeman --quiet
+  - bin/rubocop $(git ls-tree -r HEAD --name-only) --force-exclusion --format clang
+  - bin/brakeman --quiet
```
davidrunger, line 35: (Unrelated changes; I just want to be consistent in using `bin/` rather than `bundle exec`; to accomplish this, I created binstubs via `bundle binstubs rubocop` and `bundle binstubs brakeman`, as I also did with `bundle binstubs database_consistency`.)

PR #2600, opened 2020-05-19
Use Rack::Attack to ban pentesters (requesting Wordpress paths, etc)

davidrunger:
I had to switch from `dalli_store` to `mem_cache_store` to avoid a bug about "undefined method `dup_value!'".

### DIFF
--- a/config/initializers/rack_attack.rb
+++ b/config/initializers/rack_attack.rb
@@ -1,10 +1,46 @@
 # frozen_string_literal: true
 
 class Rack::Attack
+  BANNED_PATH_FRAGMENTS = Set.new(%w[
+    env
+    old-wp
+    passwd
+    php
+    phpformbuilder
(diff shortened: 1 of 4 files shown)


config/initializers/rack_attack.rb
```diff
@@ -1,10 +1,46 @@
...
   # Limit all IPs to 60 requests per clock minute
   # rubocop:disable Style/SymbolProc
   throttle('req/ip', limit: 60, period: 1.minute) do |req|
     req.ip
   end
   # rubocop:enable Style/SymbolProc
 end
+
+# Block IPs requesting Wordpress paths etc.
+# After 2 blocked requests in a day, block all requests from that IP for 1 year.
```
davidrunger, line 30: this comment is inaccurate; fixed in https://github.com/davidrunger/david_runger/pull/2601

PR #2607, opened 2020-05-20
Automatically, immediately update Rack::Attack blocked IPs as needed

davidrunger:


davidrunger:
Upon further reflection, I'm realizing that, in a multi-server-process environment (like production, where `WEB_CONCURRENCY=2`), this will only update the list of blocked IPs in a single server process.

I'm thinking (/ planning to research) about whether there's a good way to solve this problem. One thing I mean by "good" is that I'm trying to avoid a request to Redis or Memcached for every request to see if the requesting IP is blocked or not.

davidrunger:
> I'm thinking (/ planning to research) about whether there's a good way to solve this problem. One thing I mean by "good" is that I'm trying to avoid a request to Redis or Memcached for every request to see if the requesting IP is blocked or not.

I can't think of (or find) anything that will solve this problem well; I'm going to revert this.

PR #2616, opened 2020-05-21
Remove `redis` version constraint

davidrunger:
We had a version constraint when we initially added `redis` in 0ee63a5 , because that [was recommended by the `redis` README.md][1] at the time. We don't need this version constraint, though.

[1]: https://github.com/redis/redis-rb/tree/71efca9f8a40b1defd51b35890c6837e5d6032dd


davidrunger:
I removed lots of other version constraints in https://github.com/davidrunger/david_runger/pull/2617 .

PR #2635, opened 2020-05-24
Use `read_attribute_before_type_cast('created_at')` for log entries

davidrunger:
I'm hoping that this will speed up log serialization a little bit. I'm going to merge it without a performance benchmark, because I don't trust a non-performance benchmark of this, anyway.


davidrunger:
This sort undoes https://github.com/davidrunger/david_runger/commit/48cde06e8d97a9af41534d560e16a7f06ac68675 , but not really. I think that changes in the `pg` gem prevent full undoing that. I think that the `pg` gem now returns `Time`s rather than `String`s, so we can never go back fully to a pre- 48cde06e8d97a9af41534d560e16a7f06ac68675 state, I think.

PR #2661, opened 2020-05-27
Add additional metadata to `<head>` (`og:image`, etc.)

davidrunger:
The image doesn't look quite right on websites that are looking for a more horizontal image / try to coerce the image dimensions, but whatever; I don't really care. It gets the point across, as it is.

https://metatags.io/

image


davidrunger:
It looks better in Pinterest and Slack, which don't try to coerce the image dimensions:

image

PR #2662, opened 2020-05-27
Format line graph timestamp nicely and in local time in hover tooltip

davidrunger:
# Before

image

# After

image

### DIFF
--- a/app/javascript/components/charts/line_chart.js
+++ b/app/javascript/components/charts/line_chart.js
@@ -28,7 +28,13 @@ export const optionsDefaults = {
   maintainAspectRatio: false,
   responsive: true,
   scales: {
-    xAxes: [axisOptions],
+    xAxes: [{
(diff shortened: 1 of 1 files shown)


app/javascript/components/charts/line_chart.js
```diff
@@ -28,7 +28,13 @@ export const optionsDefaults = {
   maintainAspectRatio: false,
   responsive: true,
   scales: {
-    xAxes: [axisOptions],
+    xAxes: [{
+      ...axisOptions,
+      type: 'time',
+      time: {
+        tooltipFormat: 'MMMM DD YYYY, h:mma',
```
davidrunger, line 35: this uses moment.js formatting ; see `time.tooltipFormat` here https://www.chartjs.org/docs/latest/axes/cartesian/time.html and formatting options here https://momentjs.com/docs/#/displaying/ .

PR #2663, opened 2020-05-27
Validate that duration input is formatted validly

davidrunger:


app/javascript/logs/components/new_log_entry_form.vue
```diff
@@ -41,7 +43,7 @@ export default {
     inputType() {
       if (this.isText) {
         return 'textarea';
-      } else if (this.isNumber) {
+      } else if (this.isNumber || this.isCounter) {
```
davidrunger, line 46: this is an unrelated change (making sure that `counter` inputs are numbers)

PR #2684, opened 2020-05-30
Allow scheduling email if log entry hasn't been created within X time

davidrunger:


.travis.yml
```diff
@@ -38,7 +38,7 @@ script:
   - bin/rails db:schema:load
   - bin/database_consistency # needs to come after creating the DB & loading the schema
   - bin/rails immigrant:check_keys
-  - bin/annotate --models --show-indexes --sort --exclude tests && git diff --exit-code
+  - bin/annotate --models --show-indexes --sort --exclude fixtures,tests && git diff --exit-code
```
davidrunger, line 41: mostly unrelated change. this doesn't have any impact in Travis since, at this point, fixture files don't exist yet, since `spec:fixture_builder:rebuild` (the line below) hasn't been executed yet.

however, locally, if fixture files exist when this command is executed, then adding `fixtures` to the `--exclude` list will prevent annotating those fixture files (desired). thus, since this file `.travis.yml` somewhat serves as a model for commands to execute locally, I want to update this command to reflect the inclusion of this option to exclude fixtures.

config/application.rb
```diff
@@ -28,6 +28,9 @@ class DavidRunger::Application < Rails::Application
 
   # Don't generate system test files.
   config.generators.system_tests = nil
+  config.generators do |g|
+    g.factory_bot(dir: 'spec/factories')
+  end
```
davidrunger, line 33: mostly unrelated change; without this, `bin/rails g factory_bot:model LogReminderSchedule` was generating a factory definition in `test/factories/` rather than in `spec/factories/`.

app/javascript/logs/components/log_reminder_schedule_form.vue
```diff
@@ -0,0 +1,92 @@
...
+      required: true,
+    },
+  },
+};
+</script>
+
+<style>
+input {
+  width: 30px;
+}
```
davidrunger, line 91: because the stylesheet here is unscoped, this CSS rule introduced a bug, which is fixed by https://github.com/davidrunger/david_runger/pull/2692 .

PR #2693, opened 2020-05-31
Allow user to add missing phone number within groceries modal

davidrunger:
(rather than directing the user away to a separate page)


davidrunger:
I'm not going to delete the edit-user page (as I had originally thought that we'd be able to do), because it's still linked to from the "Account Settings" of the groceries app:

image

PR #2694, opened 2020-06-01
Accept phone inputs formatted in any way in api/users#update

davidrunger:
Remove non-digits and/or add 1 (the USA country code prefix) as needed.


davidrunger:
The Travis build ( https://travis-ci.org/github/davidrunger/david_runger/builds/693477181 ) passed, but GitHub doesn't seem to be registering that.

PR #2713, opened 2020-06-03
Remove Encoding::UndefinedConversionError rescue when stashing data

davidrunger:
This was added in #134 / 3297e7a .

My motivation for removing this special rescue is that the lines being deleted are untested, and I think they might not be needed anymore (based on the fact that I don't think that any of these errors have been logged to Rollbar recently).

### DIFF
--- a/app/controllers/concerns/request_recordable.rb
+++ b/app/controllers/concerns/request_recordable.rb
@@ -22,9 +22,6 @@ module RequestRecordable
         request_data.to_json,
       )
     end
-  rescue Encoding::UndefinedConversionError => error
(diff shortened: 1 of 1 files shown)


davidrunger:
^ I got code coverage to appear by reverting 739fcdfb3881b9cbe23edc0c29623b8e64fb808e , which somehow seems to break the Codecov GitHub integration (and I guess Codecov reporting in general?).

I have since removed the reversion of that commit from this PR, so I expect Codecov to stop working again.

I will follow up with a separate PR to try to fix this issue.

PR #2744, opened 2020-06-06
Set up Action Mailbox; use it to create log entries via email reply

davidrunger:


app/mailboxes/application_mailbox.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class ApplicationMailbox < ActionMailbox::Base
+  LOG_REMINDER_REPLIES_ROUTING_REGEX = /log-reminders\+(?<slug>.+)@davidrunger\.com/i.freeze
```
davidrunger, line 4: I would have preferred to define this constant within `LogReminderRepliesMailbox`, but attempting to do so was creating a circular reference that caused an error when executing `bin/database_consistency`, so I am defining the constant here instead.

app/mailers/log_reminder_mailer.rb
```diff
@@ -3,6 +3,11 @@
 class LogReminderMailer < ApplicationMailer
   def reminder(log_id)
     @log = Log.find(log_id)
-    mail(subject: %(Submit a log entry for your "#{@log.name}" log))
+    mail(
+      to: @log.user.email,
```
davidrunger, line 7: previously all of these emails were being sent to the default `to`, which was a bug. i caught it by writing specs for this mailer.

config/initializers/sidekiq.rb
```diff
@@ -2,11 +2,6 @@
 
 require 'sidekiq-scheduler' if Sidekiq.server?
 
-# rubocop:disable Security/YAMLLoad
-sidekiq_config = YAML.load(File.read('config/sidekiq.yml'))
-# rubocop:enable Security/YAMLLoad
-SIDEKIQ_QUEUES = sidekiq_config[:queues].map { |queue_name, _priority| queue_name.freeze }.freeze
```
davidrunger, line 8: Loading the Sidekiq config needed to be tweaked because we are introducing ERB into the file.

However, I then realized that we don't need to load the Sidekiq config at all, because we are only using it to get `SIDEKIQ_QUEUES`, which is a constant that we are no longer using, as of https://github.com/davidrunger/david_runger/commit/7d253ea .

config/sidekiq.yml
```diff
@@ -7,11 +7,14 @@ production:
   - ['default', 3]
   - ['rollbar', 3]
   - ['mailers', 1]
+  - ['action_mailbox_routing', 1]
 
 :schedule:
   InvalidRecordsChecker:
     cron: '55 16 * * *' # daily at 4:55pm PT
+  <% if ENV['RAILS_ENV'].in?(%w[production test]) %>
```
davidrunger, line 15: I'm doing this basically because it's annoying to have these scheduled jobs running frequently when running Sidekiq locally; it clutters the logs (and causes problems if a log statement gets printed while paused in a debugger session in Sidekiq).

config/storage.yml
```diff
@@ -0,0 +1,10 @@
+local:
+  service: Disk
+  root: <%= Rails.root.join('tmp/storage') %>
+
+amazon:
+  service: S3
+  access_key_id: <%= ENV['S3_ACCESS_KEY_ID'] %>
+  secret_access_key: <%= ENV['S3_SECRET_ACCESS_KEY'] %>
+  bucket: <%= ENV['S3_BUCKET'] %>
+  region: <%= ENV['S3_REGION'] %>
```
davidrunger, line 10: Articles that I referenced when setting this up:
1. https://medium.com/alturasoluciones/setting-up-rails-5-active-storage-with-amazon-s3-3d158cf021ff
2. https://dev.to/kylekeesling/setting-up-proper-amazon-s3-permissions-for-activestorage-3gff

spec/mailers/invalid_records_mailer_spec.rb
```diff
@@ -1,39 +1,41 @@
 # frozen_string_literal: true
 
 RSpec.describe InvalidRecordsMailer do
-  subject(:mail) { InvalidRecordsMailer.invalid_records(invalid_records_count_hash) }
+  describe '#invalid_records' do
```
davidrunger, line 4: The changes in this file are basically just whitespace changes to bring these examples within this new `describe '#invalid_records'` block.

davidrunger:
This change (which got successfully deployed via this change: https://github.com/davidrunger/david_runger/pull/2745 ) appears to be associated with a significant increase in memory usage:

image

davidrunger:
The output of `bundle exec derailed bundle:mem` (ref: https://github.com/schneems/derailed_benchmarks ) suggests that high memory usage by `action_mailbox` and/or its dependencies might be to blame:

image

davidrunger:
See https://github.com/mikel/mail/issues/ 1342 for discussion of this high memory usage and https://github.com/mikel/mail/pull/ 1343 for a possible solution.

davidrunger:
(The problem seems to be more with the `mail/parsers/address_lists_parser` dependency of `action_mailbox/engine`, rather than `action_mailbox/engine` itself.)

davidrunger:
I've merged https://github.com/davidrunger/david_runger/pull/2754 as a bandaid/"fix" for this problem. I'm bummed that I've had to go from 2 server processes down to just 1 because of this issue of excessive memory usage by `mail/parsers/address_lists_parser`, which I think that I'm barely even actually using (maybe not at all?).

PR #2747, opened 2020-06-07
Include `reply_to` field (if present) in data posted to Mailgun

davidrunger:
Relevant documentation: https://documentation.mailgun.com/en/latest/api-sending.html#sending

> **h:X-My-Header** `h:` prefix followed by an arbitrary value allows to append a custom MIME header to the message (`X-My-Header` in this case). For example, `h:Reply-To` to specify Reply-To address.

### DIFF
--- a/lib/email/mailgun_via_http.rb
+++ b/lib/email/mailgun_via_http.rb
@@ -15,11 +15,12 @@ module Email
           password: ENV['MAILGUN_API_KEY'],
         },
         body: {
-          from: mail['From'].to_s,
-          to: mail['To'].to_s,
+          from: mail.from.first,
+          to: mail.to.first,
           subject: mail.subject,
           html: mail.body.to_s,
-        },
+          'h:Reply-To' => mail.reply_to.first,
+        }.compact,
       )
     end
   end
--- a/spec/lib/email/mailgun_via_http_spec.rb
+++ b/spec/lib/email/mailgun_via_http_spec.rb
@@ -11,12 +11,16 @@ RSpec.describe Email::MailgunViaHttp do
         ::Mail::Message,
         subject: subject,
         body: email_body,
+        to: [to_email], # this stub is a bit misleading; this method doesn't return an array
+        from: [from_email], # this stub is a bit misleading; this method doesn't return an array
+        reply_to: [reply_to_email],
       )
     end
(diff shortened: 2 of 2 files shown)


lib/email/mailgun_via_http.rb
```diff
@@ -15,11 +15,12 @@ def deliver!(mail)
           password: ENV['MAILGUN_API_KEY'],
         },
         body: {
-          from: mail['From'].to_s,
-          to: mail['To'].to_s,
+          from: mail.from.first,
+          to: mail.to.first,
```
davidrunger, line 19: unrelated change; just a somewhat simpler way to access these values

spec/lib/email/mailgun_via_http_spec.rb
```diff
@@ -11,12 +11,16 @@
...
         body: email_body,
+        to: [to_email], # this stub is a bit misleading; this method doesn't return an array
+        from: [from_email], # this stub is a bit misleading; this method doesn't return an array
+        reply_to: [reply_to_email],
       )
     end
     let(:subject) { "There's a new davidrunger.com user! :) Email: davidjrunger@gmail.com." }
     let(:email_body) { 'A new user has been created with email davidjrunger@gmail.com!' }
-    let(:from_email) { 'DavidRunger.com <noreply@davidrunger.com>' }
+    let(:from_email) { '"DavidRunger.com" <reply@davidrunger.com>' }
```
davidrunger, line 21: unrelated change; just making this a little more realistic to reflect the change to our default from email here: https://github.com/davidrunger/david_runger/pull/2746/files#diff-7f4dc276bbcf65137ee281cff4ba324aR6

lib/email/mailgun_via_http.rb
davidrunger: actually, i'm realizing that this introduces a bug. for example, `mail['From'].to_s` includes the "pretty" part of an email address, whereas `mail.from.first` is just the actual email part only. so this has de-prettified our emails. i will fix this.

spec/lib/email/mailgun_via_http_spec.rb
davidrunger: Fixed in https://github.com/davidrunger/david_runger/pull/2749 .

PR #2752, opened 2020-06-07
Send pretty Reply-To field to Mailgun (not just the actual email)

davidrunger:
This fixes a bug similar to the one that we fixed for `to` and `from` in 70ca1c4 / #2749 .

### DIFF
--- a/lib/email/mailgun_via_http.rb
+++ b/lib/email/mailgun_via_http.rb
@@ -17,9 +17,9 @@ module Email
         body: {
           from: mail['From'].to_s,
           to: mail['To'].to_s,
-          subject: mail.subject,
+          subject: mail['Subject'].to_s,
           html: mail.body.to_s,
-          'h:Reply-To' => mail.reply_to.first,
+          'h:Reply-To' => mail['Reply-To'].to_s,
         }.compact,
       )
     end
--- a/spec/lib/email/mailgun_via_http_spec.rb
+++ b/spec/lib/email/mailgun_via_http_spec.rb
@@ -9,20 +9,21 @@ RSpec.describe Email::MailgunViaHttp do
     let(:mail) do
       instance_double(
         ::Mail::Message,
-        subject: subject,
+        subject: email_subject,
         body: email_body,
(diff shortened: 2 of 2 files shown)


lib/email/mailgun_via_http.rb
```diff
@@ -17,9 +17,9 @@ def deliver!(mail)
         body: {
           from: mail['From'].to_s,
           to: mail['To'].to_s,
-          subject: mail.subject,
+          subject: mail['Subject'].to_s,
```
davidrunger, line 20: this change doesn't actually make any difference, i think, but it seems like a good idea to be consistent with how we are doing this for the other fields (To, From, Reply-To). (`mail['Body']` returns `nil`, though, so we have to leave that as `mail.body.to_s`.)

lib/email/mailgun_via_http.rb
```diff
@@ -17,9 +17,9 @@ def deliver!(mail)
...
           html: mail.body.to_s,
-          'h:Reply-To' => mail.reply_to.first,
+          'h:Reply-To' => mail['Reply-To'].to_s,
```
davidrunger, line 22: this is the bugfix here

spec/lib/email/mailgun_via_http_spec.rb
```diff
@@ -9,20 +9,21 @@
...
-        to: [ugly_to_email], # this stub is a bit misleading; `#to` doesn't return an array
-        from: [ugly_from_email], # this stub is a bit misleading; `#from` doesn't return an array
-        reply_to: [reply_to_email],
+        to: [ugly_to_email],
+        from: [ugly_from_email],
+        reply_to: [ugly_reply_to_email],
       )
     end
-    let(:subject) { "There's a new davidrunger.com user! :) Email: davidjrunger@gmail.com." }
+    let(:email_subject) { "There's a new davidrunger.com user! :) Email: davidjrunger@gmail.com." }
```
davidrunger, line 19: i changed this from `subject` to `email_subject` because -- since `subject` is also a built-in RSpec method -- it seemed a little dicey to declare a `let` with that name (even though it evidently wasn't causing any problems).

PR #2754, opened 2020-06-07
Change WEB_CONCURRENCY default/fallback value to 0 (:cry:)

davidrunger:
Due to the memory increase discussed here https://github.com/davidrunger/david_runger/pull/2744#issuecomment-640219446 , we can no longer afford to run two Puma server processes.

I think that changing to 0 rather than 1 will run Puma with a single process in "single mode" rather than with 1 process in "cluser mode"; I think that "single mode" has better performance when running just 1 process.

I will also change the ENV variable value in Heroku in production to 0 to reflect this change.

### DIFF
--- a/Procfile.multi
+++ b/Procfile.multi
@@ -1,2 +1,2 @@
-web: bin/rails server -p $PORT -e $RAILS_ENV
+web: WEB_CONCURRENCY=0 bin/rails server -p $PORT -e $RAILS_ENV
 sidekiq: bin/sidekiq --config config/sidekiq.yml
--- a/config/initializers/sidekiq.rb
+++ b/config/initializers/sidekiq.rb
@@ -9,14 +9,14 @@ Sidekiq.configure_client do |config|
   # Sidekiq Client Connection Pool size:
   # We have 20 connections total from Heroku Redis Hobby Dev plan
   # ... minus 7 connections used for Sidekiq below ...
-  # ... minus 3 connections for `rails console` and general padding ...
-  # ... leaves us with 10 connections total for the Rails server processes.
-  # We are running 2 processes, so each process gets 5 connections.
-  # We'll distribute those Redis connections used by each Rails server process like so:
-  # * 2 connections for the Sidekiq client pool (the line below) (per process)
-  # * 2 for the application (per process)
-  # * 1 for Flipper (per process)
-  config.redis = ConnectionPool.new(size: 2, &build_sidekiq_redis_connection)
+  # ... minus 5 connections for `rails console` and general padding ...
+  # ... leaves us with 8 connections total for the Rails server process(es).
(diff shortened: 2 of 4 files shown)


Procfile.multi
```diff
@@ -1,2 +1,2 @@
-web: bin/rails server -p $PORT -e $RAILS_ENV
+web: WEB_CONCURRENCY=0 bin/rails server -p $PORT -e $RAILS_ENV
```
davidrunger, line 1: > I will also change the ENV variable value in Heroku in production to 0 to reflect this change.

It seems that, even after doing this, Heroku forces `WEB_CONCURRENCY` to be `1`:
```
~/workspace/david_runger  single-web-server-process-in-production ✔ 09:33:05 AM
❯ heroku config:get WEB_CONCURRENCY
0

~/workspace/david_runger  single-web-server-process-in-production ✔ 09:33:10 AM
❯ heroku run printenv | rg WEB_CONCURRENCY
Running printenv on davidrunger... up, run.5833 (Hobby)
WEB_CONCURRENCY=1
```

As a result, we are still seeing "Puma starting in cluster mode..." in the logs rather than "Puma starting in single mode...", which I think is undesirable, since I think that single mode will be faster (?). I'm hoping that setting `WEB_CONCURRENCY=0` here will override the `WEB_CONCURRENCY=1` value that Heroku seems to be trying to force on us.

config/initializers/sidekiq.rb
```diff
@@ -9,14 +9,14 @@
...
-  # * 1 for Flipper (per process)
-  config.redis = ConnectionPool.new(size: 2, &build_sidekiq_redis_connection)
+  # ... minus 5 connections for `rails console` and general padding ...
+  # ... leaves us with 8 connections total for the Rails server process(es).
+  # We are running 1 server process, so that process has 8 connections to work with connections.
+  # We'll distribute those Redis connections like so:
+  # * 3 connections for the Sidekiq client pool (the line below)
+  # * 4 for the application (in config/initializers/redis.rb)
+  # * 1 for Flipper (in config/initializers/flipper.rb)
+  config.redis = ConnectionPool.new(size: 3, &build_sidekiq_redis_connection)
```
davidrunger, line 19: Now that we are only running a single server process, we have more Redis connections to distribute to that one process (vs before, when we had to split the Redis connections between two processes).

Procfile.multi
davidrunger: It did indeed work; the app booted in single mode:
```
Jun 07 09:47:15 davidrunger app/web.1 16:47:15 web.1     | Puma starting in single mode...
```

davidrunger:
This change seems to have succeeded in sufficiently lowering the application's memory usage:

image

In fact, our memory usage appears to be lower than ever in recent history (with the downside being, of course, that we now have fewer processes(/threads) to handle web requests).

PR #2761, opened 2020-06-08
Add basic feature test for logs app

davidrunger:


app/javascript/shared/customized_vue.js
```diff
@@ -73,7 +73,7 @@ if (window.davidrunger.env !== 'test') {
 Vue.use(Vuex);
 
 Vue.use(VueForm);
-if (window.davidrunger.env !== 'test') {
+if (window.location.pathname !== '/packs-test/mocha_runner.html') {
```
davidrunger, line 76: we need this `if` to be truthy when running feature specs (so that the Vue router gets initialized and added to Vue components and whatnot), but falsy when running JavaScript unit tests (since it becomes problematic if those tests try to interact with / change the path of the browser runner).

prior to this change, the `if` was falsy in both cases (since `window.davidrunger.env === 'test'` in both cases).

this change makes the conditional behave as desired (described above).

PR #2762, opened 2020-06-08
Add basic feature test for workout app

davidrunger:


app/javascript/workout/workouts_table.vue
```diff
@@ -12,7 +12,7 @@ table(v-if='workouts.length')
       td {{workout.created_at | prettyTime}}
       td(v-if='!isOwnWorkouts') {{workout.username}}
       td {{(workout.time_in_seconds / 60).toFixed(1)}}
-      td {{JSON.stringify(workout.rep_totals).replace(/{|}|"/g, '').replace(/,/g, ', ')}}
```
davidrunger, line 15: I noticed that we were formatting rep totals like `chinups:100, pushups:300` (without spaces between the colon and the number). i changed the formatting to `chinups: 100, pushups: 300` (with spaces). this lengthened the line of code, so I extracted it into a filter (which is probably a good idea, anyway).

PR #2764, opened 2020-06-08
Add test for UsersController#update failure case

davidrunger:


Gemfile
```diff
@@ -89,6 +89,7 @@ group :test do
   gem 'json-schema'
   gem 'launchy' # for save_and_open_page in feature specs
   gem 'percy-capybara'
+  gem 'rails-controller-testing'
```
davidrunger, line 92: this gives us the `render_template` matcher

PR #2768, opened 2020-06-08
Add test for unauthorized Api::UsersController#update attempt

davidrunger:


spec/spec_helper.rb
```diff
@@ -113,6 +113,15 @@
     allow(Rails).to receive(:cache).and_return(ActiveSupport::Cache::MemCacheStore.new)
   end
 
+  config.define_derived_metadata(file_path: %r{/spec/controllers/api/}) do |metadata|
+    # we leverage this metadata for the `config.before(:each, requset_format: :json)` setting below
+    metadata[:requset_format] = :json
+  end
+
+  config.before(:each, requset_format: :json) do
+    request.accept = 'application/json'
```
davidrunger, line 122: I guess that the point of this change was probably to make it so that we enter the `format.json` block [here](https://github.com/davidrunger/david_runger/blob/be5d2d33a847e7cbcf329a3dc22f40bd6d2b4667/app/controllers/application_controller.rb#L76-L78)?

PR #2769, opened 2020-06-08
Add test for Logs::UploadsController#index

davidrunger:
fixes https://rollbar.com/davidjrunger/davidrunger/items/241/

Adding this test surfaced the fact that this controller action has actually been 500ing (since this change  https://github.com/davidrunger/david_runger/commit/852cd4112ee0c141f52e0b7620cb5a9591f42a24#diff-a68281def00c6c6e8ca731b5f1b1727fR14 ). This PR fixes the bug introduced by that change (linked above) in addition to adding test coverage to ensure that we won't have a regression.

### DIFF
--- a/app/views/layouts/application.haml
+++ b/app/views/layouts/application.haml
@@ -11,7 +11,7 @@
     %meta{property: 'og:title', content: title}
 
     -# other metadata
-    - if current_page?(controller: 'home', action: 'index')
+    - if current_page?(controller: '/home', action: 'index')
       %meta{name: 'author', content: 'David Runger'}
       %meta{property: 'og:type', content: 'website'}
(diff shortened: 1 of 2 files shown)


app/views/layouts/application.haml
```diff
@@ -11,7 +11,7 @@
     %meta{property: 'og:title', content: title}
 
     -# other metadata
-    - if current_page?(controller: 'home', action: 'index')
+    - if current_page?(controller: '/home', action: 'index')
```
davidrunger, line 14: this is a bugfix to prevent Rails from interpreting this as a relative controller (`logs/home`) when accessing a nested controller (i.e. `logs/uploads`, in the relevant case surfaced in this PR). adding a `/` at the beginning of the controller name causes Rails to interpret the controller name "absolutely".

PR #2779, opened 2020-06-09
Color rails server log lines red and append 2 newlines in development

davidrunger:
This makes it way easier to see each request separately, rather than the requests all bleeding together in a continuous stream of log lines without any separation between them.

# Before

image

# After

image

**_I know which one I prefer_!** 👍 😄 IMO the latter format makes it _way_ easier to distinguish each separate request, at a glance.

### DIFF
--- a/lib/david_runger/log_formatter.rb
+++ b/lib/david_runger/log_formatter.rb
@@ -14,6 +14,7 @@ class DavidRunger::LogFormatter < Lograge::Formatters::KeyValue
   end
 
   def call
-    super(@data) # use the Lograge::Formatters::KeyValue#call method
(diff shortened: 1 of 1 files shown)


lib/david_runger/log_formatter.rb
```diff
@@ -14,6 +14,7 @@ def initialize(data)
   end
 
   def call
-    super(@data) # use the Lograge::Formatters::KeyValue#call method
+    log_message = super(@data)
+    Rails.env.development? ? "#{log_message.red}\n\n\n\n" : log_message
```
davidrunger, line 18: (the String#red method is provided in development by `amazing_print`)

PR #2781, opened 2020-06-09
Upload code coverage result to Codecov using bash uploader

davidrunger:
(This is in addition to the built-in/automatic uploading that will also be done by the `SimpleCov::Formatter::Codecov` Ruby formatter at the completion of the `bin/rails spec:rb` task.)

This is part of an effort to debug the issue(s) discussed here: https://community.codecov.io/t/500-unknown-error/1525/ .


davidrunger:
The issue has been resolved! 🙌 Closing.

PR #2783, opened 2020-06-09
Log full request path, including query string

davidrunger:
For whatever reason, `lograge` [removes][1] the query string from the logged data. I prefer to see the query string as part of the path. (The query string params do show up in the `params` that are all logged, but I we don't know, for example, if a param came from a query string or from a POST request body, which might be worth knowing.)

[1]: https://github.com/roidrage/lograge/blob/1729eab/lib/lograge/log_subscribers/action_controller.rb#L31

### DIFF
--- a/lib/david_runger/log_builder.rb
+++ b/lib/david_runger/log_builder.rb
@@ -10,6 +10,7 @@ class DavidRunger::LogBuilder
 
   def extra_logged_data
     {
+      path: request.fullpath,
       exception: exception_log,
       params: params_log,
       status: status_code_log,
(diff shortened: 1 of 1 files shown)


lib/david_runger/log_builder.rb
```diff
@@ -10,6 +10,7 @@ def initialize(event)
 
   def extra_logged_data
     {
+      path: request.fullpath,
```
davidrunger, line 13: here we are not adding a new attribute/key to this hash, but rather this will override the `path` attribute set by `lograge`, which [strips the query string](https://github.com/roidrage/lograge/blob/1729eab/lib/lograge/log_subscribers/action_controller.rb#L31). (thank you `lograge` for giving precedence to my override.)

PR #2785, opened 2020-06-09
Change default Capybara driver to Chrome headless

davidrunger:
The vast majority of our feature specs (all but one, I think) use Chrome headless (via the `:js` spec metadata/tag), so let's just make it the default and opt out in the rare cases where we don't want to use Chrome.

Specs that need to use the `:rack_test` driver may opt in via the `:rack_test_driver` spec metadata flag that this PR adds support for.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -35,7 +35,7 @@ Capybara.register_driver(:chrome_headless) do |app|
   options = Selenium::WebDriver::Chrome::Options.new(args: %w[no-sandbox headless disable-gpu])
   Capybara::Selenium::Driver.new(app, browser: :chrome, options: options)
 end
-Capybara.default_driver = :rack_test
+Capybara.default_driver = :chrome_headless
 Capybara.javascript_driver = :chrome_headless
 # allow loading JS & CSS assets via `save_and_open_page` when running `rails s`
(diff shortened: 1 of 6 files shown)


spec/spec_helper.rb
```diff
@@ -35,7 +35,7 @@
   options = Selenium::WebDriver::Chrome::Options.new(args: %w[no-sandbox headless disable-gpu])
   Capybara::Selenium::Driver.new(app, browser: :chrome, options: options)
 end
-Capybara.default_driver = :rack_test
+Capybara.default_driver = :chrome_headless
```
davidrunger, line 38: here's the main change of this PR

spec/spec_helper.rb
```diff
@@ -35,7 +35,7 @@
...
 Capybara.javascript_driver = :chrome_headless
```
davidrunger, line 39: (note that I'm also leaving `:chrome_headless` as the `Capybara.javascript_driver` because, otherwise, a spec where someone adds the `:js` flag will try to use something called `geckdriver` and will fail. so by leaving this setting, we'll run Chrome whether or not we have a `:js` metadata flag for a spec. the only way to use a different driver is to opt in to it manually, e.g. with a `:rack_test_driver` metadata flag.)

PR #2825, opened 2020-06-11
Only run brakeman if Ruby or HAML files have been changed

davidrunger:


bin/orchestrate-tests.rb
```diff
@@ -283,9 +288,14 @@ def required_tasks(target_tasks, known_dependencies: [], trimmable_requirements:
...
 
+    memoize \
+    def haml_files_changed?
+      (Dir['app/**/*.haml'] & files_changed).any?
+    end
+
     memoize \
     def ruby_files_changed?
-      (Dir['**/*.{rb}'] & files_changed).any?
+      (Dir['**/*.rb'] & files_changed).any?
```
davidrunger, line 298: unrelated change here; i noticed that we only have one extension here, so the curly braces aren't necessary

PR #2826, opened 2020-06-11
Only run JavaScript tests if JavaScript/Vue files have changed

davidrunger:
I'm thinking about deleting the JavaScript tests altogether. For now, though, this (running them only when JavaScript/Vue files have changed) should help, maybe enough that it'll be worth keeping the JavaScript unit tests.

### DIFF
--- a/bin/orchestrate-tests.rb
+++ b/bin/orchestrate-tests.rb
@@ -237,6 +237,7 @@ class OrchestrateTests < Pallets::Workflow
     RunDatabaseConsistency => proc { !OrchestrateTests.db_schema_changed? },
     RunEslint => proc { !OrchestrateTests.files_with_js_changed? },
     RunImmigrant => proc { !OrchestrateTests.db_schema_changed? },
+    RunJsSpecs => proc { !OrchestrateTests.files_with_js_changed? },
     RunRubocop => proc { !ruby_files_changed? },
     RunStylelint => proc { !OrchestrateTests.files_with_css_changed? },
     SetupJs => proc do |tentative_list|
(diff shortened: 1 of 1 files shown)


bin/orchestrate-tests.rb
```diff
@@ -304,12 +307,26 @@ def ruby_files_changed?
...
+        trimmable_requirements << task
+        new_tentative_requirements.delete(task)
+      end
     end
-  true_requirements = required_tasks(target_tasks, trimmable_requirements: trimmable_requirements)
+    true_requirements = new_tentative_requirements
+    true_requirements = required_tasks(true_requirements, trimmable_requirements: trimmable_requirements)
+
+    break if true_requirements_at_start.map(&:name).sort == true_requirements.map(&:name).sort
+  end
```
davidrunger, line 328: this code is horrifying. there is almost certainly a cleaner/better way to do what i'm doing here. for now, though, my goal is just to get this working correctly, which it seems to be doing. (some tests would probably be wise, though, to verify that (now and going forward)...)

davidrunger:
This is our biggest speed improvement so far:

https://travis-ci.org/github/davidrunger/david_runger/builds

image

That's because when we don't run JavaScript tests, we can skip [this step](https://travis-ci.org/github/davidrunger/david_runger/builds/697106216) (which I've excerpted from the logs for a different branch's previous build):

```
'bin/webpack --silent' succeeded (exited with 0, took 38.217).
2020-06-11T05:22:08.7586Z pid=8694 wid=W1928 wfid=POTEF4A3B2D773 jid=JSJS5696254890 wf=OrchestrateTests tsk=SetupJs INFO: Done
```

That step is one of our two webpack compilation steps, the other compilation step being when we build the JavaScript assets for our Ruby tests. One webpack compilation is much faster than two. :)

PR #2827, opened 2020-06-11
Split Travis env variables onto separate lines via `global` config

davidrunger:
I have tried to do this previously, which -- without the `global` option -- triggered Travis's "build matrix" functionality, which was not what I want. I just want to trigger a single build, which is what (I think) this accomplishes, while also allowing the variables to be split onto separate lines, for better clarity/readability.

I saw that this `global` config option is available when looking at these docs https://docs.travis-ci.com/user/speeding-up-the-build/#rspec-parallelization-example for an unrelated reason.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,11 @@ addons:
   chrome: stable
   postgresql: "9.6"
 env:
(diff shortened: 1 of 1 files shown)


davidrunger:
Part of the reason that I wanted to make this change is because it just touches one YAML file -- no Ruby, JavaScript, Vue, CSS, HAML, etc files -- and so we skip many of our tests/checks (per many PRs that I've made today).

The result is a really fast build time (relatively speaking to our historical build times):

https://travis-ci.org/github/davidrunger/david_runger/builds/697121209

image

About the first 90 seconds of that is just setting up the environment (Ruby, node, yarn, Postgres, npm packages, gems, Redis, Chrome, etc...); it's only the last 44 seconds or so that is actually spent running the test suite. Not bad, if you ask me!

PR #2831, opened 2020-06-12
Delete `.reek.yml` config file

davidrunger:
I don't actively use reek. Every once in a while I do use it, but I'm thinking it will be a while before I use it again on this codebase, if ever. Therefore, let's clean up the root directory a bit by deleting this file.

The main motivation for this PR is that I'm curious to see how fast the Travis CI build time will be for this branch build that just changes a `.yml` file. :) I have been making a lot of efforts lately to speed up the CI build time.


davidrunger:
> The main motivation for this PR is that I'm curious to see how fast the Travis CI build time will be for this branch build that just changes a .yml file. :) I have been making a lot of efforts lately to speed up the CI build time.

Results:
1. 1 min 50 seconds, then I rebiult that build and got
2. 1 min 53 seconds ( https://travis-ci.org/github/davidrunger/david_runger/builds/697440432 )

These results should be understood in light of the fact that the first ~1 minute of the build time is Travis setting up the environment. So the part of the build time that we control is only taking ~50 seconds for this build, which is pretty good.

Previously, the Travis setup time took closer to 90 seconds, I think (because we were using Ruby, node, and yarn versions not available on Travis by default and so which we had to spend time to install), and the overall build time was closed to maybe 3 minutes and 18 seconds or so, which gives 108 seconds of runtime by our build process (198 seconds - 90 seconds).

In that perspective, the build time for this branch is 50 seconds vs a 108 seconds baseline, which is just 46% as long as before, i.e. the part of the build that is under our control can now run more than twice as fast for a minimal changeset such as this.

PR #2845, opened 2020-06-12
Don't add `yarn` to PATH explicitly

davidrunger:


davidrunger:
I created this PR in the hope that Travis is already adding `yarn` to the PATH for us / putting `yarn` in the PATH. Happily, based on the fact that the build passed, this does seem to be the case. 🎉 

My motivation for doing this is that now all four `Check[...]Version` build steps have no prerequisite; I will combine them into a single step in a future PR.

PR #2865, opened 2020-06-15
Allow inclusion of a note when creating a log entry via an auth_token

davidrunger:


davidrunger:
Actually, I don't think that we really want this feature.

For tracking Travis build time (the motivation that I originally had for this feature, so that the list of which test tasks had been executed could be included as a note), I think it will be better to allow Travis to post to `Api::LogEntriesController#create` rather than going though `LogsController#index`. Closing.

davidrunger:
>  I think it will be better to allow Travis to post to `Api::LogEntriesController#create`

See https://github.com/davidrunger/david_runger/pull/2866 for this feature.

PR #2882, opened 2020-06-16
Bump active_actions from 0.3.0 to 0.5.1 ; add shaped as a dependency

davidrunger:
(We are installing `shaped` explicitly because it is needed by `active_actions` but cannot be installed automatically by `active_actions` because `shaped` is only released via GitHub.)

Also, define the expected shape of `message_params` hashes using `Shaped::Hash` This is a new feature (provided via `shaped`) that is included in our recent bump of `active_actions` from 0.3.0 to 0.5.1 .

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,11 +3,6 @@
 source 'https://rubygems.org'
 ruby '2.7.0'
 
-git_source(:github) do |repo_name|
-  repo_name = "#{repo_name}/#{repo_name}" unless repo_name.include?('/')
-  "https://github.com/#{repo_name}.git"
-end
-
 gem 'active_actions', github: 'davidrunger/active_actions'
 gem 'active_model_serializers'
 gem 'administrate'
@@ -40,6 +35,7 @@ gem 'rack-mini-profiler', require: false
 gem 'rails', github: 'rails/rails'
 gem 'redis'
 gem 'rollbar'
+gem 'shaped', github: 'davidrunger/shaped'
 gem 'sidekiq'
 gem 'sidekiq-scheduler', require: false # required manually in config/initializers/sidekiq.rb
 gem 'stackprof' # Provides stack traces for flamegraph for rack-mini-profiler.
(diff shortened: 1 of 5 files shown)


Gemfile
```diff
@@ -3,11 +3,6 @@
 source 'https://rubygems.org'
 ruby '2.7.0'
 
-git_source(:github) do |repo_name|
-  repo_name = "#{repo_name}/#{repo_name}" unless repo_name.include?('/')
-  "https://github.com/#{repo_name}.git"
-end
```
davidrunger, line 9: I think that this is provided automatically now by RubyGems or something; things still seem to work after deleting it.

app/controllers/api/text_messages_controller.rb
```diff
@@ -9,7 +9,10 @@ def create
       SmsRecords::SendMessage.new(
         user: current_user,
         message_type: text_message_params['message_type'],
-        message_params: text_message_params['message_params'].to_unsafe_h,
+        message_params:
+          text_message_params['message_params'].
+            to_unsafe_h.
+            transform_values { Integer(_1) },
```
davidrunger, line 15: somewhat strangely, although the test setup has `store_id: user.stores.first.id` (where that value should be an `Integer`), when we reach this point in the application code (during tests), the value is a string.

in the real application flow, though, it is an Integer.

I guess that there is some thing in the Rails testing code/flow that stringifies any integer values in the params, since that sometimes probably happens in application code.

anyway, this workaround to explicitly re-integer-ize the value works, i guess.

davidrunger: i was able to fix it by adding `as: :json` to the `post` request in tests, instead. thanks to https://stackoverflow.com/a/36350469/4009384 and https://stackoverflow.com/questions/36333434/rails-rspec-json-integers-being-converted-to-strings-when-testing-post-call#comment77547488_36350469 for this solution. :) 👍

PR #2887, opened 2020-06-17
Decrease max test task time to 60 seconds

davidrunger:
I don't think that there are any tasks that we expect to take this long. Let's find out sooner rather than later if any tasks do take this long.


davidrunger:
Actually, it seems that the job doesn't get _aborted_ if it reaches the timeout limit; the job is _retried_ (see documentation comment here: https://github.com/linkyndy/pallets/blob/b93ea47/lib/pallets/configuration.rb#L23-L25 ). (See Travis build log for evidence of this: https://travis-ci.org/github/davidrunger/david_runger/builds/699170675 .) This is not really the behavior that we want.

I guess that I'll just delete this config, then, and leave the default job timeout period of [30 minutes](https://github.com/linkyndy/pallets/blob/b93ea47/lib/pallets/configuration.rb#L56).

davidrunger:
> I guess that I'll just delete this config, then, and leave the default job timeout period of 30 minutes.

See https://github.com/davidrunger/david_runger/pull/2888

PR #2895, opened 2020-06-17
Automatically apply `as: :json` option for API controller tests

davidrunger:
If `as: :json` is _not_ set, then Rails's testing flow (more specifically, [`rails-controller-testing`](https://github.com/rails/rails-controller-testing), I guess?) will convert number params (e.g. `store_id: 123`) to strings (e.g. `'store_id' => '123'`). This isn't realistic for most/all API requests, since those requests are made with JSON data payloads, which _can_ support actual numbers (unlike non-API requests where e.g. the params are encoded in the path or query params, and therefore are strings and cannot be actual numbers).

Therefore, this PR changes the default behavior for all simulated requests in API controller tests to automatically add the `as: :json` option.

### DIFF
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -39,4 +39,11 @@ Rails.application.configure do
   # config.action_view.raise_on_missing_translations = true
 
   config.active_storage.service = :local
+
+  extra_load_paths = [
+    Rails.root.join('spec/support'),
+  ].map(&:to_s).map(&:freeze)
+  config.eager_load_paths.concat(extra_load_paths)
 end
+
+Rails.autoloaders.main.do_not_eager_load(Rails.root.join('spec/support/matchers/'))
--- a/spec/controllers/api/text_messages_controller_spec.rb
+++ b/spec/controllers/api/text_messages_controller_spec.rb
@@ -6,7 +6,7 @@ RSpec.describe Api::TextMessagesController do
   let(:user) { users(:user) }
 
   describe '#create' do
-    subject(:post_create) { post(:create, params: params, as: :json) }
+    subject(:post_create) { post(:create, params: params) }
 
     let(:valid_params) do
       {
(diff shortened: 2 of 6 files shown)


config/environments/test.rb
```diff
@@ -39,4 +39,9 @@
   # config.action_view.raise_on_missing_translations = true
 
   config.active_storage.service = :local
+
+  extra_load_paths = [
+    Rails.root.join('spec/support'),
+  ].map(&:to_s).map(&:freeze)
+  config.eager_load_paths.concat(extra_load_paths)
```
davidrunger, line 46: Without doing this, zeitwerk was not automatically defining a `Monkeypatches` module for `Monkeypatches::MakeAllRequestsAsJson` to live in.

spec/controllers/api/text_messages_controller_spec.rb
```diff
@@ -6,7 +6,7 @@
   let(:user) { users(:user) }
 
   describe '#create' do
-    subject(:post_create) { post(:create, params: params, as: :json) }
```
davidrunger, line 9: being able to remove this manual `as: :json` specification here (and not having to add it to all other requests made in API controller tests) is the purpose of this PR.

config/environments/test.rb
```diff
@@ -39,4 +39,11 @@
...
 end
+
+Rails.autoloaders.main.do_not_eager_load(Rails.root.join('spec/support/matchers/'))
```
davidrunger, line 49: This is because the files in `spec/support/matchers/` don't follow `zeitwerk` conventions; for example, [`spec/support/matchers/match_schema.rb`](https://github.com/davidrunger/david_runger/blob/be5d2d33a847e7cbcf329a3dc22f40bd6d2b4667/spec/support/matchers/match_schema.rb) does not define a `Matchers::MatchSchema` constant.

PR #2896, opened 2020-06-17
Install `rails-controller-testing` from GitHub master branch

davidrunger:
I think that there might have been some recent-ish changes to fix Ruby 2.7 warnings about keyword arguments. The 1.0.4 version that we were sourcing via RubyGems was released way back on December 05, 2018.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -88,7 +88,7 @@ group :test do
   gem 'json-schema'
   gem 'launchy' # for save_and_open_page in feature specs
   gem 'percy-capybara'
-  gem 'rails-controller-testing'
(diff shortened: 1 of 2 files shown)


davidrunger:
> I think that there might have been some recent-ish changes to fix Ruby 2.7 warnings about keyword arguments.

Yes, this significantly reduces the number of hits for `Using the last argument as keyword parameters is deprecated` in test logs, from [10](https://travis-ci.org/github/davidrunger/david_runger/builds/699197840) to [3](https://travis-ci.org/github/davidrunger/david_runger/builds/699198618). 🙌

PR #2897, opened 2020-06-17
Set up a separate database for each of the 4 RSpec-running test tasks

davidrunger:
When tests are executing in parallel and all using the same database, I think that there is a risk that some of the concurrent attempts to change the database can run afoul of each other. I don't think that I've ever seen this as a problem in any Travis runs, but I do think that I have seen it locally.

I also think that there is somewhat of a risk that the database transactions might block/delay each other, when run concurrently using the same database?

Therefore, by creating a separate DB for each RSpec test-running task (RunFeatureTests, RunApiControllerTests, RunHtmlControllerTests, RunUnitTests), I am hoping that we can:
1. increase parallelism / decrease CI build time
2. eliminate the risk of flakiness caused by concurrent attempts to modify a single database

### DIFF
--- a/bin/test/requirements_resolver.rb
+++ b/bin/test/requirements_resolver.rb
@@ -20,6 +20,7 @@ class Test::RequirementsResolver
         Test::Tasks::CompileJavaScript => Test::Tasks::YarnInstall,
         Test::Tasks::SetupDb => nil,
         Test::Tasks::BuildFixtures => Test::Tasks::SetupDb,
+        Test::Tasks::CreateDbCopies => Test::Tasks::BuildFixtures,
 
         # Checks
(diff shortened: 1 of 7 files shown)


davidrunger:
That was way, way easier than I expected...

Let's merge it and see what happens! So far, it doesn't seem to be speeding things up much, but it's somewhat hard to tell since rubocop is running for this branch (versus some other PRs e.g. dependency updates where rubocop usually doesn't run).

davidrunger:
Recent build times ( Google login required: https://www.davidrunger.com/users/1/logs/travis-build-times ) suggest that this might have actually slowed things down a bit (maybe even up to ~10 seconds). Even if that's true, though, this still seems worthwhile to me, to avoid the risk of flakiness caused by our concurrent test execution?

PR #2902, opened 2020-06-17
Remove grocery items and stores from admin

davidrunger:
I don't really care about these types of records very much. Since nobody actually uses davidrunger.com for their grocery list, all of this data is fake.

Removing these leaves Users, Requests, and IpBlocks in admin, which are really the only things that I care about viewing via the admin interface.


config/routes.rb
```diff
@@ -46,9 +46,7 @@
 
   namespace :admin do
     resources :users
-    resources :items
     resources :requests, only: %i[index show]
-    resources :stores
```
davidrunger, line 51: It took me the longest time to figure out why links to "Items" and "Stores" were still appearing in the admin side nav after I had deleted everything else. Opening up the `administrate` source code (`bundle open administrate`) quickly revealed that it was because I had forgotten to remove these routes. I'm glad that it reminded me! The more deletions, the merrier.

PR #2905, opened 2020-06-17
Add controller tests (#index & #show) for `Admin::RequestsController`

davidrunger:
Also, make the `Request` factory a bit more interesting and randomized.

### DIFF
--- a/spec/factories/requests.rb
+++ b/spec/factories/requests.rb
@@ -31,29 +31,24 @@
 #
 
 FactoryBot.define do
-  chrome_user_agent = <<-CHROME.squish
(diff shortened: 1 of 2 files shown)


spec/factories/requests.rb
```diff
@@ -31,29 +31,24 @@
...
     format { 'html' }
     status { 200 }
-    view { 12 }
-    db { 8 }
-    ip { '77.88.47.71' }
-    user_agent { chrome_user_agent }
-    requested_at { 1.week.ago }
-  end
-
-  trait :chrome do
```
davidrunger, line 56: we aren't using this factory trait anywhere anymore. (last/only usage was deleted here: https://github.com/davidrunger/david_runger/pull/1017/files#diff-c673dd26f8a6b35b73cbd0b15e1c928eL7 .)

PR #2909, opened 2020-06-17
Add specs for ApplicationPolicy

davidrunger:


app/policies/application_policy.rb
```diff
@@ -49,7 +49,7 @@ def initialize(user, scope)
     end
 
     def resolve
-      scope
+      scope.none
```
davidrunger, line 52: Writing these specs surfaced this bug, which is that our default `ApplicationPolicy` scope included _all_ records, rather than _no_ records. We want the default to be non-permissive, hence this change.

PR #2912, opened 2020-06-17
Add tests for SaveRequest worker

davidrunger:
These specs are nasty (I hate stubbing stuff in specs that we don't need to stub), but it's part of the journey that I am taking toward 100% test coverage.


davidrunger:
Too slow, again, Percy.

PR #2923, opened 2020-06-18
Lock `redis` gem to `~> 4.1.4`

davidrunger:
This is because there are some annoying warnings in 4.2.0 and greater that haven't been addressed by Sidekiq yet.

Relatedly, we're also removing the no-longer-needed monkeypatch of `Kernel#warn` that was suppressing the warnings from `redis` >=4.2.0.


davidrunger:
The biggest motivation that I have for making this change (locking the `redis` gem and removing our monkeypatch of `Kernel::warn`) is largely that the monkeypatch was difficult(/impossible?) to write test(s) for.

PR #2925, opened 2020-06-18
Add Rack::Attack tests

davidrunger:


spec/spec_helper.rb
```diff
@@ -106,14 +106,11 @@
     end
   end
 
-  # rubocop:disable RSpec/HookArgument
-  config.before(:each) do
-    Rack::Attack.reset!
```
davidrunger, line 111: I think that `Rails.cache` is a `ActiveSupport::Cache::NullStore` in most specs, so I don't really think that this `    Rack::Attack.reset!` call was doing anything in those cases.

Additionally, when `Rails.cache` is a `ActiveSupport::Cache::MemCacheStore`, calling `Rack::Attack.reset!` raises an error about that cache store not implementing some method.

That's why I'm deleting this `Rack::Attack.reset!` call.

We are [calling `Rails.cache.clear`](https://github.com/davidrunger/david_runger/pull/2925/files#diff-6b64bcaea6cadff14a786eb2be83ea69R8) in the `around` block for the specs being added in this PR. That should be sufficient.

spec/spec_helper.rb
```diff
@@ -106,14 +106,11 @@
...
-  end
-  # rubocop:enable RSpec/HookArgument
-
-  config.before(:each, :cache) do
-    allow(Rails).to receive(:cache).and_return(ActiveSupport::Cache::MemCacheStore.new)
```
davidrunger, line 116: I changed this from a `before` block to an `around` because I think that the `before` block was executing "too late", i.e. after the `around` block in `rack_attack_spec.rb`. Switching this to an `around` block seems to cause it to be invoked before the `around` block in `rack_attack_spec.rb`, which is good; that way, `Rails.cache` will be an instance of `ActiveSupport::Cache::MemCacheStore.new` when we reach the `around` block in `rack_attack_spec.rb`, which is what we want.

PR #2931, opened 2020-06-18
Delete development-only middlewares (for server and Sidekiq)

davidrunger:


davidrunger:
That's what I'm talking about, baby! "100%" "code coverage" (for Ruby files in `app/`, `config/`, and `lib/`)!

image

Unfortunately, that metric is tainted by the fact that we have a _few_ lines that are ignored via `:nocov:` comments (although really not many; I think fewer than 10 lines are thusly ignored).

The other thing is that "code coverage" just means "the line was hit at least once", which is very far from "the code is tested in a way that verifies that the code has the intended behavior".

davidrunger:
Here's another celebratory screenshot:

https://codecov.io/gh/davidrunger/david_runger

image

The bad thing is that the only direction that code coverage can go from here is down. 😛 I'll feel pressure to keep the code coverage at 100% for future PRs.

PR #2941, opened 2020-06-19
Revert "Bump redis from 4.1.4 to 4.2.1"

davidrunger:
Reverts davidrunger/david_runger#2933

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -35,7 +35,7 @@ gem 'rack-mini-profiler', require: false
 gem 'rails', github: 'rails/rails'
 # We can probably remove this version constraint once Sidekiq 6.1 has been released.
 # See https://github.com/mperham/sidekiq/issues/4591 .
-gem 'redis', '~> 4.2.1'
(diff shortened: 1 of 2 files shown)


davidrunger:
This reverts a PR that accidentally undid this PR: https://github.com/davidrunger/david_runger/pull/2923 . See motivation there. tl;dr we are logging this a ton:

> Jun 19 08:40:39 davidrunger app/web.1 15:40:38 sidekiq.1 | `Redis#exists(key)` will return an Integer in redis-rb 4.3. `exists?` returns a boolean, you should use it instead. To opt-in to the new behavior now you can set Redis.exists_returns_integer =  true. To disable this message and keep the current (boolean) behaviour of 'exists' you can set `Redis.exists_returns_integer = false`, but this option will be removed in 5.0. (/app/vendor/bundle/ruby/2.7.0/gems/sidekiq-6.0.7/lib/sidekiq/launcher.rb:160:in `block (2 levels) in ❤')

PR #2956, opened 2020-06-21
Disable NewRelic JavaScript tracking

davidrunger:
It doesn't seem to be providing any useful data, and I think it throws some errors in the console and makes Firefox warn about privacy.

image

### DIFF
--- a/config/newrelic.yml
+++ b/config/newrelic.yml
@@ -27,6 +27,10 @@ common: &default_settings
   distributed_tracing:
     enabled: true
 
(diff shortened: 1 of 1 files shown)


davidrunger:
# After

Much better!!!

image

davidrunger:
Those two remaining errors in the console log (both talking about the same issue) are caused by my Firefox adblocker addon, uBlock Origin. Disabling the addon for davidrunger.com makes those errors not appear in the console after refreshing the page.

PR #2957, opened 2020-06-21
Add Request#auth_token association & populate it when auth token used

davidrunger:


app/controllers/concerns/request_recordable.rb
```diff
@@ -22,20 +20,6 @@ def store_initial_request_data_in_redis
...
-    rescue StashRequestError => error
-      cause_error = error.cause
-      Rails.logger.warn(<<-LOG.squish)
-        Failed to store request data in redis,
-        error=#{error.class}: #{error.message},
-        cause=#{cause_error&.class}: #{cause_error&.message},
-        request_data=#{request_data.inspect}
-      LOG
-      Rollbar.warning(error, request_data: request_data)
-    end
```
davidrunger, line 38: This code wasn't getting hit anymore, I think because we are not stubbing `request.request_id` in all controller specs. I can't really think how we might hit this code (to write a test for it), so let's just delete this `rescue`.

PR #2959, opened 2020-06-21
Swith HTTP client from HTTParty to Faraday

davidrunger:
I realized that HTTParty is making requests with `Content-Type: x-www-form-urlencoded` rather than `Content-Type: application/json`, and this bothers me.

I also realized that the "User-Agent" for requests made by HTTParty is just "Ruby", which is also a bit lame (very generic).

I also realized that we are already requiring `faraday` implicitly as a dependency of `oauth2`.

Faraday didn't end up working out quite as well as I would have liked, but at the end of the day I think it's a bit of an improvement.


Gemfile
```diff
@@ -13,12 +13,13 @@ gem 'connection_pool'
 gem 'dalli'
 gem 'devise'
 gem 'email_reply_trimmer', github: 'discourse/email_reply_trimmer'
+gem 'faraday'
+gem 'faraday_middleware'
```
davidrunger, line 17: this gives us `conn.request(:json)` and `conn.response(:json)`

app/actions/sms_records/save_sms_record.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class SmsRecords::SaveSmsRecord < ApplicationAction
-  requires :nexmo_response, HTTParty::Response
+  requires :nexmo_response, Faraday::Response
```
davidrunger, line 4: I guess that we should probably have `SmsRecords::PostToNexmo` be responsible for parsing this into a `Hash` so that `SmsRecords::SaveSmsRecord` doesn't have to change when we change our HTTP library (i.e. this would be `requires :nexmo_response, Hash`, or, better yet, instead of `Hash` a description of the required/expected shape of the hash).

davidrunger: (I made this change.)

davidrunger:
Because we are posting to the Travis build log as JSON now (rather than `x-www-form-urlencoded`), we see that the data param is correctly being provided/parsed as a number now.

# Before

image

# After

image

PR #2964, opened 2020-06-22
Add `after_action :verify_authorized` to `ApplicationController`

davidrunger:


app/controllers/api/text_messages_controller.rb
```diff
@@ -1,8 +1,6 @@
 # frozen_string_literal: true
 
 class Api::TextMessagesController < ApplicationController
-  after_action :verify_authorized
```
davidrunger, line 4: (deleting `after_action :verify_authorized` from this particular controller because we now call it in `ApplicationController`)

app/controllers/api/logs_controller.rb
```diff
@@ -43,7 +44,7 @@ def log_params
   end
 
   def set_log
-    @log ||= current_user.logs.find_by(id: params['id'])
+    @log = current_user.logs.find_by(id: params[:id])
```
davidrunger, line 47: unrelated change from `'id'` to `:id` for consistency

app/models/item.rb
```diff
@@ -18,6 +18,7 @@
 
 class Item < ApplicationRecord
   belongs_to :store
+  has_one :user, through: :store
```
davidrunger, line 21: this facilltates `@record.user == @user` in the `ApplicationPolicy` by giving the record a `#user` method when the record is an Item

app/policies/auth_token_policy.rb
```diff
@@ -1,15 +1,4 @@
...
-    true
-  end
-
-  def destroy?
-    @record.user == @user
-  end
-
-  def update?
-    @record.user == @user
-  end
```
davidrunger, line 14: the `ApplicationPolicy` now reflects these defaults, so there's no need to override anymore

app/policies/workout_policy.rb
```diff
@@ -1,12 +1,7 @@
 # frozen_string_literal: true
 
 class WorkoutPolicy < ApplicationPolicy
-  class Scope
-    def initialize(user, scope)
-      @user = user
-      @scope = scope
-    end
```
davidrunger, line 8: this initialize method does the same thing as the initialize method of `ApplicationPolicy::Scope`, which we are now inheriting from

app/views/layouts/application.haml
```diff
@@ -11,7 +11,8 @@
     %meta{property: 'og:title', content: title}
 
     -# other metadata
-    - if current_page?(controller: '/home', action: 'index')
+    -# (note: the curly braces are necessary to avoid a warning about keyword params)
+    - if current_page?({ controller: '/home', action: 'index' })
```
davidrunger, line 15: unrelated change to avoid Rails deprecation warning

PR #2984, opened 2020-06-23
Set up Action Cable and use it to live-update logs with new entries

davidrunger:


app/javascript/logs/components/log.vue
```diff
@@ -45,7 +45,7 @@ div
           v-model='publiclyViewable'
           @change='savePubliclyViewableChange'
         ) Publicly viewable
-      div(v-if='!publiclyViewable')
+      div(v-if='isOwnLog && !publiclyViewable')
```
davidrunger, line 48: unrelated bugfix for an issue that prevented a user from viewing another user's log that had been shared with their email specifically but not made publicly visible

config/application.rb
```diff
@@ -3,12 +3,12 @@
 require_relative 'boot'
 
 require 'rails'
+require 'action_cable/engine'
 require 'action_controller/railtie'
 require 'action_mailbox/engine'
 require 'action_mailer/railtie'
 require 'action_view/railtie'
 require 'active_job/railtie'
-require 'active_model/railtie'
```
davidrunger, line 11: unrelated change; i think that `active_model` already is being required by someone else (maybe `active_record`) so this isn't really necessary so i am removing it

app/javascript/logs/components/log.vue
davidrunger: test covering the bugfix added here: https://github.com/davidrunger/david_runger/pull/2998

PR #3057, opened 2020-07-02
Use `runger_style` gem for spec/, db/migrate/, bin/ rubocop configs

davidrunger:


davidrunger:
Moved to https://github.com/davidrunger/david_runger/pull/3065 . Closing.

PR #3088, opened 2020-07-04
Delete extra word "connections" from comment about Redis connections

davidrunger:
This is a typo resulting from some careless changes to this comment in f5e3794 .


config/initializers/sidekiq.rb
```diff
@@ -8,10 +8,10 @@
 Sidekiq.configure_client do |config|
   # Sidekiq Client Connection Pool size:
   # We have 20 connections total from Heroku Redis Hobby Dev plan
-  # ... minus 7 connections used for Sidekiq below ...
+  # ... minus 7 connections used for the Sidekiq server below ...
```
davidrunger, line 11: this is not a typo fix; this is an unrelated clarification that I am squeezing in

PR #3096, opened 2020-07-04
Check for `AWS_*` env vars (not `S3_*`) when booting production

davidrunger:
fixes https://rollbar.com/davidjrunger/davidrunger/items/276/

This should have been changed as part of #3092 / 7c6b91c .

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -113,9 +113,9 @@ Rails.application.configure do
   require_relative('../../lib/email/mailgun_via_http.rb')
   config.action_mailer.delivery_method = Email::MailgunViaHttp
 
-  if %w[S3_ACCESS_KEY_ID S3_SECRET_ACCESS_KEY S3_BUCKET S3_REGION].all? { ENV[_1].present? }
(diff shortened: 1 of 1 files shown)


config/environments/production.rb
```diff
@@ -113,9 +113,9 @@
   require_relative('../../lib/email/mailgun_via_http.rb')
   config.action_mailer.delivery_method = Email::MailgunViaHttp
 
-  if %w[S3_ACCESS_KEY_ID S3_SECRET_ACCESS_KEY S3_BUCKET S3_REGION].all? { ENV[_1].present? }
+  if %w[AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY S3_BUCKET S3_REGION].all? { ENV[_1].present? }
     config.active_storage.service = :amazon
   else
-    puts('Skipping :amazon storage enablement because not all required ENV variables are present.')
+    raise(':amazon storage cannot be enabled because not all required ENV variables are present')
```
davidrunger, line 119: The issue being fixed by this PR would have been easier to debug if production had failed to boot (good).

Had production failed to boot, then that would have broken all of production rather than just breaking Active Storage (bad).

That's a trade-off that I'm willing to make, though.

Hopefully it won't come up again either way.

PR #3387, opened 2020-08-30
Don't look for auth token with a blank secret

davidrunger:
Doing this early `return` saves a database query (that we expect to always return `nil`).


davidrunger:
Because my davidrunger.com domain name is expired/expiring, tests "failed". I am going to merge anyway.

PR #3388, opened 2020-08-31
Use `cloudflare-rails` for correct IP reporting via Cloudflare

davidrunger:


davidrunger:
I guess we'll have to just merge this and then test if it works in production.

PR #3398, opened 2020-08-31
Loosen IP validation format of `IpBlock` for IPv6

davidrunger:
fixes https://rollbar.com/davidjrunger/davidrunger/items/318/

### DIFF
--- a/app/actions/ip_blocks/store_request_block_in_redis.rb
+++ b/app/actions/ip_blocks/store_request_block_in_redis.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class IpBlocks::StoreRequestBlockInRedis < ApplicationAction
-  requires :ip, String, format: { with: /[.:0-9a-z]{7,39}/ }
+  requires :ip, String, format: { with: /\A[.:0-9a-f]{7,39}\z/ }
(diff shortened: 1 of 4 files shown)


app/actions/ip_blocks/store_request_block_in_redis.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class IpBlocks::StoreRequestBlockInRedis < ApplicationAction
-  requires :ip, String, format: { with: /[.:0-9a-z]{7,39}/ }
+  requires :ip, String, format: { with: /\A[.:0-9a-f]{7,39}\z/ }
```
davidrunger, line 4: somewhat unrelated change; I am just updating this regex. I should probably extract this to a constant or something to keep it DRY (and reference that constant wherever we need an IPv6 regex), but whatever.

davidrunger: i think that there might be a gem that already has some of these validation regexes

davidrunger: maybe i am thinking of this one: https://github.com/mdespuits/validates_formatting_of

PR #3885, opened 2020-12-22
Add tests for ability to disable any Sidekiq worker via Flipper flag

davidrunger:


davidrunger:
This brings our test coverage back to 100%. 😄

PR #3910, opened 2020-12-31
Fetch 100 commits for GitHub Actions

davidrunger:
Without this, everything is fetched as a single commit, which was then breaking our ability to determine which files have been changed since `master` (which, in turn, we use to determine which tests/linters should be run).

### DIFF
--- a/bin/test/diff_helpers.rb
+++ b/bin/test/diff_helpers.rb
@@ -25,7 +25,7 @@ module DiffHelpers
   def files_changed
     if !system('git log -1 --pretty="%H" master > /dev/null 2>&1')
       puts('`master` branch is not present; fetching it now...')
-      `git fetch origin master:master --depth=1`
+      system('git fetch origin master:master --depth=1', exception: true)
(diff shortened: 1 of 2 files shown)


davidrunger:
Evidence for claim "Without this [change to our workflow setup], everything is fetched as a single commit":

> Fetches only a single commit by default

-- https://github.com/actions/checkout

bin/test/diff_helpers.rb
```diff
@@ -25,7 +25,7 @@ def diff_mentions_rubocop?
   def files_changed
     if !system('git log -1 --pretty="%H" master > /dev/null 2>&1')
       puts('`master` branch is not present; fetching it now...')
-      `git fetch origin master:master --depth=1`
+      system('git fetch origin master:master --depth=1', exception: true)
```
davidrunger, line 28: Unrelated change, but in investigating this issue, I was wondering if this command was failing (though it wasn't) and so changed it to raise an error if it does fail.

PR #3912, opened 2021-01-01
Switch from `administrate` to `activeadmin`

davidrunger:
We added `administrate` in this commit: https://github.com/davidrunger/david_runger/commit/1c658b1657693565c8eb4409570d72467be487ee#diff-ceddb52394452ed1594eac9f8db36fef9f05cadec806e7e9108eafa16e6c936b

However, I've found it to be a little bit less flexible than I would like. For example, I want to add [`devise_masquerade`](https://github.com/oivoodoo/devise_masquerade), which requires a line like this in a view: `= link_to "Login As", masquerade_path(user)`. However, `administrate` doesn't even have/use views, exactly; instead, it offers the option to write HTML in a "field" definition. However, we aren't really talking about a "field" in this case; we just want to add an extra link somewhere to impersonate a user.

Another limitation that I haven't appreciated about `administrate` is that there isn't much support for filtering the index views.

Thus, we are switching in this PR to using `activeadmin` as our admin framework, since I know that it does a better job of addressing the above two concerns.


app/controllers/concerns/request_recordable.rb
```diff
@@ -13,6 +13,8 @@ module RequestRecordable
     before_action :store_initial_request_data_in_redis
   end
 
+  private
```
davidrunger, line 16: unrelated change; I noticed that this method is public but there is no good reason for it to be

app/models/user.rb
```diff
@@ -34,6 +32,8 @@ class User < ApplicationRecord
   has_many :text_log_entries, through: :logs
   has_many :workouts, dependent: :destroy
 
+  devise
```
davidrunger, line 35: I am just moving this down because we'll probably have occasion to read `devise` far less often than we'll want to check the `User` validations and/or associations.

PR #3913, opened 2021-01-01
Consolidate to just webpack-cli@3.3.12 (not also 4.3.0)

davidrunger:
This allows the `bin/webpack-dev-server` command to work; otherwise it doesn't work (though I don't remember the exact error).

This largely undoes 2fc0c0b (Add webpack-cli).

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,6 @@
     "vuex-router-sync": "^5.0.0",
     "waypoints": "^4.0.1",
     "webpack": "^4.44.2",
-    "webpack-cli": "^4.3.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -74,7 +74,6 @@
     "vuex-router-sync": "^5.0.0",
     "waypoints": "^4.0.1",
     "webpack": "^4.44.2",
-    "webpack-cli": "^4.3.0",
```
davidrunger, line 77: By removing this `webpack-cli` declaration from our own `package.json`, we rely on only whichever `webpack-cli` version `webpacker` depends upon.

PR #3916, opened 2021-01-03
Perform `User` Google omniauth without using Devise `:omniauthable`

davidrunger:
Motivation: We want to add an `AdminUser` model that will (like `User`) authenticate via Google OAuth. However, Devise only allows one model to declare `:omniauthable`, and declaring `:omniauthable` for any one model essentially makes it impossible to do manual Google OAuth for a different model (https://github.com/heartcombo/devise/wiki/OmniAuth-with-multiple-models). This change, therefore, paves the way for us to add an `AdminUser` model that will also authenticate via Google OAuth.


spec/features/google_login_spec.rb
```diff
@@ -9,7 +9,8 @@
   end
 
   before do
-    OmniAuth.config.mock_auth[:google_oauth2,] = OmniAuth::AuthHash.new(
+    OmniAuth.config.add_mock(
+      :google_oauth2,
```
davidrunger, line 13: this is an unsubstantive change (just more concise)

